PR #10, opened 2023-08-24
[kokrui] iP

kokrui:
# a heading

- a bullet list
- a bullet list

1. a numbered list 
2. a numbered list 

```python
print("a fenced code block with syntax highlighting")
```

- [x] a task list
- [ ] wow task

🅰️n emoji

> a block quote

[a hyperlink](https://github.com/nus-cs2103-AY2324S1/ip/pull/10)

some `inline` code

**bold** *italics* ~~strikethrough~~


src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+     * Constructs a new Task with the given name and sets its completion status to false.
+     *
+     * @param name The name of the task.
+     */
+    public Task(String name) {
+        this.name = name;
+        this.isDone = false;
+    }
+
+    public static Task fromFileFormat(String line) throws DukeException {
```
ruishanteo: Should this method have a javadoc?

src/main/java/Event.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Parses the command string to create an Event instance.
+     *
+     * @param input The command string.
+     * @return A new Event instance.
+     * @throws DukeException If the input format is invalid.
+     */
+    public static Event createFromCommandString(String input) throws DukeException {
```
ruishanteo, line 30: Will you be adding time on top of dates, as per Level 8?

src/main/java/FileUtils.java
```diff
@@ -0,0 +1,49 @@
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class FileUtils {
```
ruishanteo: I think this class is missing header comments, as well as method documentations.

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Returns a string representation of the task to be displayed in a Todo List.
+     *
+     * @return A string representation of the task.
+     */
+    @Override
+    public String toString() {
+        return "[" + (isDone ? "X" : " ") + "] " + this.name;
```
ruishanteo, line 57: Maybe you can consider using String.format for this.

ruishanteo (commented):
Overall I find your code easy to read and executed very well!

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
+public abstract class Task {
```
ruishanteo: Should all files be in packages?

src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+public class DukeException extends Exception {
```
ruishanteo: Header comments would be good to have for public classes.

ruishanteo (commented):
All the files look good, except for some missing documentation!

src/main/java/Task.java
kokrui: Thanks! I agree that it should have additional documentation due to its complexity and I've added the JavaDoc

src/main/java/FileUtils.java
kokrui: Thanks! I agree that it should have a header comment since it's a public class so I've added it

src/main/java/Task.java
kokrui: Thanks! I agree that all classes should belong to packages so I've since made that change

src/main/java/DukeException.java
kokrui: Thanks! I agree that it should have a header comment since it's a public class so I've added it

PR #15, opened 2023-08-24
[sopa301] iP

sopa301:
# My Commit Straight From the Trash Can

It's figurative trash so don't look at it. Now that's out of the way, the rest of the text is to meet the requirements.

Here's a bullet list:
- One
- Two
- Three

Here's a numbered list:
1. One
2. Two
3. Three

Here's a `Java` code block:
```Java
function void foo() {
    System.out.println("whoops.");
}
```

Here's a task list:
- [X] Enroll in CS.
- [ ] Graduate from CS.
- [ ] Get a job.

Here's an emoji: 🤡 

Here's a quote:
> “Why waste time say lot word when few word do trick?”  -Kevin Malone

Here's a [hyperlink](https://www.youtube.com/watch?v=dQw4w9WgXcQ).

Here's some `inline code`.

Here's some **bold**, *italic* and ~~strikethrough~~ text.


src/main/java/bot/Bot.java
```diff
@@ -0,0 +1,84 @@
...
+    public String greet() {
+        return ui.showWelcome();
+    }
+
+    /**
+     * Indicates if the bot is active.
+     *
+     * @return True if the bot is active, else false.
+     */
+    public boolean isExit() {
```
ruiyangzh: "Names representing methods must be verbs and written in camelCase."
 Perhaps a method name such as getExitStatus?

src/main/java/bot/utils/Command.java
```diff
@@ -0,0 +1,382 @@
...
+    public static Command find(String str) {
+        return new FindCommand(str);
+    }
+
+    /**
+     * Checks if the bot should exit after the execution of the command.
+     *
+     * @return True if the bot should exit, false otherwise.
+     */
+    public abstract boolean isExit();
```
ruiyangzh: As noted.

src/main/java/bot/utils/TaskList.java
```diff
@@ -0,0 +1,129 @@
...
+    public int size() {
+        return this.list.size();
+    }
+
+    /**
+     * Gets an iterator of the tasks in the list.
+     *
+     * @return Iterator of Task objects.
+     */
+    public Iterator<Task> iterator() {
```
ruiyangzh: Perhaps getIterator?

ruiyangzh (commented):
Hi, not much to comment on, well done generally. (Outside of the scope of this code review, but maybe would be more compartmentalized if you split the child Command classes into their own class files?)

PR #16, opened 2023-08-24
Finished Level0

oeggy03:
1. Named the bot eggbot
2. Finished level 0: Implement an initial skeletal version of the it that simply greets the user and exits.


oeggy03:
Merged with main

PR #20, opened 2023-08-24
[KamiliArsyad] iP

KamiliArsyad:
## Meowies

> “Meow meow meowies” – My Cat ([source](thecatsite.com))

Meowies meows for your meowness. Check my website [kamiliarsyad.com](kamiliarsyad.com) to find out more!

- Cat-based
- Interactive

All you need to do is, download it from [here](https://kamiliarsyad.com).
- [x] double-click it.
- [ ] add your tasks.
- [ ] let it manage your tasks for you 😉
- [ ] And it is FREE!

Features:

1. Meows
2. Managing tasks
3. Managing deadlines
4. Reminders

In case you are curious about a box of code, here is one to satisfy your curiosity.
```java
public class Main {
        System.out.println("meows");
    }
}
```


src/main/java/duke/UI.java
```diff
@@ -0,0 +1,55 @@
...
+
+        public void bye() {
+            printer("Bye. Hope to see you again soon!");
+        }
+
+        public void echo(String input) {
+            printer(input);
+        }
+
+        private void printer(String input) {
```
thaddeusong: Probably better to use the verb "print" as the method name here.

src/main/java/duke/task/EventTask.java
```diff
@@ -0,0 +1,83 @@
...
+        this.dateEnd = LocalDate.parse(dateEnd);
+    }
+
+    @Override
+    protected String getTaskTypeString() {
+        return squareBracketWrapper("E");
+    }
+
+    @Override
+    protected String saveToFileString() {
```
thaddeusong, line 48: This method sounds like it saves to the file, when it just returns the string in a save format. Maybe a name like toSaveString or toSaveFormat would be better?

src/main/java/duke/task/TaskStorage.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        this.tasks.add(task);
+        this.saveToFile();
+
+        return "added: " + task;
+    }
+
+    /**
+     * Mark a task as done.
```
dhruvir29: The first sentence of the JavaDoc comment should start with Returns, Sends, Adds instead of Return, Send, Add. Perhaps change this to "Marks a task as done".

src/main/java/duke/task/TaskStorage.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public String markAsDone(int index) {
+        this.tasks.get(index).markAsDone();
+        this.saveToFile();
+        return "Nice! I've meowrked this task as done:\n"
+                + "    " + this.tasks.get(index);
+    }
+
+    /**
+     * Unmark a task as done.
```
dhruvir29: Similarly, consider changing this to "Unmarks a task as done".

src/main/java/duke/task/TaskStorage.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public String unmarkAsDone(int index) {
+        this.tasks.get(index).unmarkAsDone();
+        this.saveToFile();
+        return "Oh meow! I've marked this task as undone :( :\n"
+                + "    " + this.tasks.get(index);
+    }
+
+    /**
+     * Delete a task.
```
dhruvir29: Similarly, "Deletes a task".

src/main/java/duke/task/TaskStorage.java
```diff
@@ -0,0 +1,178 @@
...
+        Task task = this.tasks.get(index);
+        this.tasks.remove(index);
+        this.saveToFile();
+        return "Noted. I've removed this task:\n"
+                + "    " + task + "\n"
+                + "    " + "Now you have " + this.tasks.size() + " tasks in the list.";
+    }
+
+    /**
+     * Find tasks that contain the keyword.
```
dhruvir29: Similarly, "Finds tasks that contain the keyword"

src/main/java/duke/InputHandler.java
```diff
@@ -0,0 +1,61 @@
+package duke;
+
+import java.util.Scanner;
+
+import duke.task.TaskStorage;
+
+class InputHandler {
```
dhruvir29: Perhaps add a header comment for the class InputHandler as it seems to be a requirement.

PR #38, opened 2023-08-25
[Rongzhi] iP

rngzhi:
# Jarvis Chatbot
> "Jarvis, sometimes you gotta run before you can walk." - Tony Stark

## Description
The name **Jarvis** stands for [Just A Rather Very Intelligent System](https://en.wikipedia.org/wiki/J.A.R.V.I.S.). This pull request adds task management capabilities to the Jarvis chatbot, allowing users to 
* Create tasks
* List tasks
* Mark tasks as completed

**Jarvis** currently supports three kinds of tasks:
*1. ToDo: without time component*
*2. Deadline: with a due time*
*3. Event: consists of a start time(from) and an end time(to)*

The task management feature enhances Jarvis's utility by enabling users to keep track of their to-do list without leaving the chat interface.

## Features
- [x] Task Management
- [ ] Check tasks with given date(coming in next update)
- [ ] Store tasks in disk(coming in next update)

## How to Test
1. Install the Jarvis chatbot on your preferred platform.
2. Run `Jarvis` file to launch the chatbot 🥰

___
### If you are a Java programmer, you can also use it to practice Java. Here's the `main` method:
```java
    Jarvis chatbot = new Jarvis("records.txt");
    chatbot.run();
}
```


src/main/java/ContentMissingException.java
```diff
@@ -0,0 +1,14 @@
+public class ContentMissingException extends JarvisException{
```
RoeReRe, line 1: Missing JavaDoc all many classes?

src/main/java/Jarvis.java
```diff
@@ -0,0 +1,59 @@
...
+                break;
+            }
+            horizontal_line_printer();
+            System.out.println(input);
+            horizontal_line_printer();
+        }
+        sc.close();
+    }
+
+    public static void manage_tasks(List<Task> taskList) {
```
RoeReRe, line 50: Perhaps its better to camel case your method names?

src/main/java/TaskManagement.java
```diff
@@ -0,0 +1,183 @@
...
+            System.out.println(" Noted. I've removed this task:");
+            System.out.println("\t" + target.toString());
+            Jarvis.horizontal_line_printer();
+
+        } catch (IndexOutOfBoundsException e) {
+            throw new InvalidTaskIndexException(index);
+        }
+    }
+
+    private void add_task(String input) throws ContentMissingException, InvalidCommandException {
```
RoeReRe, line 119: Would just wrapping these under JarvisException be an issue?

src/main/java/JarvisException.java
```diff
@@ -0,0 +1,6 @@
+public class JarvisException extends Exception{
```
RoeReRe, line 1: I like how you extended from JarvisException to differentiate the different types of errors.

src/main/java/JarvisException.java
```diff
@@ -0,0 +1,6 @@
...
+    @Override
+    public String getMessage() {
+        return "OOPS!!! ";
```
RoeReRe, line 4: Perhaps leaving the whitespace for the inherited classes would be clearer? One would have to be cognisant that the parent class already has a whitespace; and this isn't consistent with how other Java exceptions are implemented.

```suggestion
        return "OOPS!!!";
```

RoeReRe (commented):
Hi! I'm assigned your repo for the 2nd PR review for iP.

I think your code is quite nicely implemented! However, I think Javadoc is mandatory for the iP and we have to be quite careful with how we name methods or risk penalization.

All the best!

src/main/java/ContentMissingException.java
coderhuang559, line 1: Perhaps MissingContentException sounds smoother?

src/main/java/TaskManagement.java
```diff
@@ -0,0 +1,183 @@
...
+                }
+            } catch (ContentMissingException | InvalidTaskIndexException
+                     | InvalidCommandException e) {
+                System.out.println(e.getMessage());
+            }
+        }
+        sc.close();
+    }
+
+    private void list_printer() {
```
coderhuang559, line 62: Maybe camelCase your method names. There are a few other instances as well.

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+
+
+
+    public Task(String content) {
+        this.content = content;
+        this.isDone = false;
+    }
+
+    public void mark() {
+        this.isDone = true;
```
coderhuang559, line 13: isDone is a good boolean name. Good job! 😁

src/main/java/TaskManagement.java
```diff
@@ -0,0 +1,183 @@
+import java.security.Key;
+import java.util.List;
+import java.util.Scanner;
+
+public class TaskManagement {
+    private List<Task> taskList;
+    enum Keyword {
```
coderhuang559, line 7: Creative way to handle user commands!

src/main/java/TaskManagement.java
```diff
@@ -0,0 +1,183 @@
...
+            return keyword;
+        }
+    }
+
+
+    public TaskManagement(List<Task> taskList) {
+        this.taskList = taskList;
+    }
+
+    public void operate() {
```
coderhuang559, line 28: Very clean code in the `operate` method. Impressive!

coderhuang559 (commented):
Overall good attempt. Some details to fix.

PR #50, opened 2023-08-26
[SimWPEric] iP

SimWPEric:
### DukePro

> “Your mind is for having ideas, not holding them.” – David Allen [source](https://dansilvestre.com/productivity-quotes)

- text-based
* easy to learn
+ ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/nus-cs2103-AY2324S1/ip/pull/50).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :stuck_out_tongue_winking_eye:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---
```java
public class Main {
    }
}
```


src/main/java/duke/MainWindow.java
```diff
@@ -0,0 +1,78 @@
...
+    private void handleUserInput() {
+        String input = userInput.getText();
+        String response = duke.getResponse(input);
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(input, userImage),
+                DialogBox.getDukeDialog(response, dukeImage)
+        );
+
+        // Closes the GUI
+        if (input.equals("bye")) {
```
simbayippy: what if user input is "BYE"?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Parses user input and generates the appropriate command for Duke.
+     *
+     * @param input The user's input command.
+     * @return The corresponding Command object based on the user input.
+     * @throws DukeInvalidArgumentException If the input does not match any valid command format.
+     * @throws DukeInvalidDateFormatException If the input contains an invalid date format.
+     * @throws DukeEndDateBeforeStartDateException If an event's end date is before its start date.
+     */
+    public static Command parseUserInput(String input) throws DukeInvalidArgumentException,
```
simbayippy, line 190: similarly what is users input are not all lowercase?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,156 @@
...
+            if (!dukeFile.createNewFile()) {
+                throw new IOException("Failed to create duke file");
+            }
+        }
+    }
+
+    private void writeTaskListToFile(TaskList taskList) throws IOException {
+        Path filePath = Paths.get(this.filepath);
+        try {
+            Files.deleteIfExists(filePath);
```
simbayippy, line 135: is this unnecessary? when doing File file = new File(filePath); i believe it should already override the existing file in the filepath

simbayippy (commented):
Overall your code is brilliantly done. Abstraction of classes are done very clearly. Took away some learning points from your code too - thank you!

src/main/java/duke/MainWindow.java
SimWPEric: Thank you for pointing this out, I have added `input.toLowerCase()` to make the exit command case insensitive.

src/main/java/duke/parser/Parser.java
SimWPEric: Yes, you are right, I have added additional format specifiers ie: `String todoPattern = "(?i)todo (.+)";`, where `(?i)` makes the command case insensitive when parsing user input.

src/main/java/duke/storage/Storage.java
SimWPEric: My idea is to overwrite the existing file rather than appending to an existing file. Please let me know if you have any alternative implementations !

PR #54, opened 2023-08-26
[qyaner] iP

qyaner:
# Duke 

> “Your mind is for having ideas, not holding them.” – David Allen

Duke frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~FAST~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/qyaner/ip).
2. double-click it.
3. add your tasks.
4.let it manage your tasks for you 😉

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

```ruby
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
+import duke.*;
```
sk2001git: I think you should not use the * operator for imports, it is generally recommended that you import every package individually, i think i see this at some other class files as well

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,99 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a task with a deadline.
+ *
+ * @author Qin Yan Er
+ */
+public class Deadline extends Task {
+    protected String by;
```
sk2001git: I think a noun would be better to describe the String "by"!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,135 @@
...
+        System.out.println(list.getTask(number).toString());
+    }
+
+    /**
+     * Prints when user unmarks a done task back to not done yet.
+     *
+     * @param list the list of tasks.
+     * @param number the task number to be unmarked.
+     */
+    public void unMark(TaskList list, int number) {
```
sk2001git: I think this should be unmark since its is a word ?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param list the list of tasks.
+     * @param newEvent the new event to be added.
+     */
+    public void addEvent(TaskList list, Event newEvent) {
+        System.out.println("Got it. I've added this task:");
+        System.out.println(newEvent.toString());
+        System.out.println("Now you have " + list.getSize() + " tasks in the list.");
+    }
+
```
sk2001git: In general, I think that you have done quite well in documentation, naming of methods and variable names, naming seem intuitive and well ordered, well done! 👍 😄

src/main/java/duke/Deadline.java
sk2001git: i think a possible suggestion could be  "startTime" ?

sk2001git (commented):
LGTM. I think there was some minor coding standard violations and naming problems, so small number of nits however naming was generally well done! 👍

PR #67, opened 2023-08-26
[Wern] iP

wxwern:
# Todoify

> "The best CLI task manager" - nobody

Todoify is easy to use. It:
- has a chatbot-like interface
- available via your preferred [DE](https://en.wikipedia.org/wiki/Desktop_environment) or even your [terminal](https://en.wikipedia.org/wiki/Computer_terminal).
- uses Unix- and POSIX-like CLI syntax, so experienced users can pick it up quickly.
- ~~FAST~~ _SUPER-FAST_ to use with just your keyboard.

## Usage

1. Check out the [releases](https://github.com/wxwern/ip/releases).
2. Download it.
3. Run the JAR file by double-clicking it, or using `java -jar <jar file name>`.
4. Talk to the chatbot to add your tasks!

## Features

- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [ ] Adds more RAM
- [ ] Makes your computer slower

## Implementation

It uses common patterns like **Object-Oriented Programming** and **Observer Pattern**, which makes the source code clean, easy to interact with, and easy to extend. For instance, here's a snippet of how the designed class structures allow for **integration into *any* user-interface** of your liking in seconds:

```java
public class Main {
 
        // Example to demonstrate processing a text-only interface with the chatbot. 
        // Notice that this is easily translatable into almost any form of GUI.

        // Initialise with default settings.
        Chatbot chatbot = new Chatbot(null, null);

        // Add a listener for new messages and output them.
        chatbot.addEventListener(message -> {
            if (message.getSenderType() == ChatMessage.SenderType.CHATBOT) {
                System.out.printf("[%s]: ", chatbot.getName());
                System.out.println(message.getMessage());
                System.out.println();
            }
        });

        // Start processing standard input to feed into the chatbot.

        chatbot.openConversation();
        while (chatbot.isConversationOpen()) {
            System.out.print("[You]: ");
            System.out.flush();

            if (!scanner.hasNextLine()) {
                break;
            }

            String line = scanner.nextLine();
            System.out.println();

            chatbot.sendMessageFromUser(line);
        }

        chatbot.closeConversation();
        scanner.close();
    }
}
```


src/main/java/todoify/MainCli.java
```diff
@@ -0,0 +1,62 @@
...
+ * The chatbot CLI entry-point, which provides an interactive input prompt via the CLI.
+ *
+ * <p>
+ * The default name for the chatbot is specified at {@link todoify.chatbot.Chatbot#DEFAULT_NAME}. Custom names can be
+ * set by modifying the {@link MainCli#main} implementation here.
+ * </p>
+ */
+public class MainCli {
+
+    private static final String CHATBOT_PADDING = "  ";
```
Dioclei: Well done, you did not make the common mistake of writing constants with a non-fully-capitalised name.

src/main/java/todoify/storage/InternalStorage.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * A class that manages the program's internal storage's read-write operations.
+ */
+public class InternalStorage {
+
+    private static InternalStorage sharedInstance = null;
```
Dioclei: Why does the InternalStorage class need a shared instance among all InternalStorage classes? I feel that this naming could be confusing to readers, perhaps a more descriptive name could help.

src/main/java/todoify/taskmanager/TaskManager.java
```diff
@@ -0,0 +1,198 @@
...
+        try {
+            String data = this.storageHandler.loadFrom(this.storageLocation);
+            JsonArray array = JsonParser.parseString(data).getAsJsonArray();
+
+            // Prepare a new list of tasks.
+            List<Task> tasks = new ArrayList<>();
+
+            // Prepare a new set of classes, from most specific to least specific.
+            // This ordering is required to match the provided JSON to a class that's as specific as possible.
+            @SuppressWarnings("unchecked")
```
Dioclei: Why do you use @SuppressWarnings here? Perhaps you could write an explanation in the header comment!

src/main/java/todoify/util/EpochConverter.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public class EpochConverter {
+
+    /**
+     * Converts the given date-time string to number representing seconds since the Unix epoch.
+     *
+     * @param dateTimeString The date time string to convert, in ISO format of "yyyy-MM-ddThh:mm:ss".
+     * @return The number of seconds since the Unix epoch of 1970-01-01 00:00:00 UTC.
+     * @throws DateTimeParseException if the date-time cannot be parsed.
+     */
```
Dioclei, line 26: This is a great header comment!

Dioclei (commented):
I feel that your code is really well done! Just a few small suggestions here and there, but honestly great job! 👍 👍 👍

src/main/java/todoify/chatbot/ChatCommand.java
```diff
@@ -0,0 +1,270 @@
...
+                currParamKey = part;
+                currParamValue = new StringBuilder();
+            } else {
+                currParamValue.append(parts[i]);
+                currParamValue.append(' ');
+            }
+        }
+        if (!currParamKey.isEmpty()) {
+            params.put(currParamKey, currParamValue.toString().trim());
+        }
```
victorlaiyeeteng: The method `parse` seems a little long. Is it possible to factor out this part of the function to improve readability of the function?

src/main/java/todoify/chatbot/ChatCommand.java
```diff
@@ -0,0 +1,270 @@
...
+        case "bye":
+        case "exit":
+            return this.type = Operation.EXIT;
+
+        default:
+            break;
+        }
+
+        return this.type = Operation.UNKNOWN;
+    }
```
victorlaiyeeteng: I like the use of enums to improve the understanding of the function.

src/main/java/todoify/storage/InternalPath.java
```diff
@@ -0,0 +1,101 @@
...
+        return Arrays.copyOf(this.components, this.components.length);
+    }
+
+    /**
+     * Returns a new path relative to the configured base directory, while removing the last component in the path.
+     *
+     * @return The path after removing the last path component.
+     * @throws NegativeArraySizeException if there are no more components in the path to strip.
+     */
+    public InternalPath excludingLastComponent() {
```
victorlaiyeeteng:
```suggestion
    public InternalPath excludeLastComponent() {
```

src/main/java/todoify/storage/InternalPath.java
```diff
@@ -0,0 +1,101 @@
...
+        return new InternalPath(pathComponents);
+    }
+
+    /**
+     * Returns a new path relative to the configured base directory, while adding the given last component to the path
+     *
+     * @param component The new component to append to the end of the path.
+     * @return The path after adding the new last path component.
+     */
+    public InternalPath appendingLastComponent(String component) {
```
victorlaiyeeteng:
```suggestion
    public InternalPath appendLastComponent(String component) {
```

src/main/java/todoify/taskmanager/task/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.deadline;
+    }
+
+    @Override
+    public String toString() {
+        return String.format(
+                "<D> %s %s (by: %s)",
+                this.getCompletedIndicatorString(),
+                this.getTitle(),
+                this.deadline == null ? null : EpochConverter.getUserReadableStringFromEpoch(this.getDeadline())
```
victorlaiyeeteng: Perhaps, you can abstract this part of processing the deadline output in a separate function?

src/test/java/todoify/chatbot/ChatCommandTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        command = ChatCommand.parse(" todo blabla\tblah blah\n123\n");
+        assertEquals("todo", command.getName());
+        assertEquals("blabla blah blah 123", command.getData());
+        assertEquals(ChatCommand.Operation.ADD_TODO, command.getOperation());
+        assertFalse(command.hasParams());
+        assertFalse(command.hasParam("blah"));
+        assertFalse(command.hasParamWithUsefulValue("blah"));
+        assertNull(command.getParam("blah"));
+    }
```
victorlaiyeeteng, line 45: I like the elaborate testing done on the ChatCommand objects.

victorlaiyeeteng (commented):
Overall, I like the Code Quality of your code, but there are a few suggestions that could perhaps improve the readability of some functions.

src/main/java/todoify/taskmanager/TaskManager.java
wxwern: Thanks for the note, I forgot to include that in the comment. The reason was arrays cannot be initialised with generics, so in this particular case:
```java
Class<Task>[] availClasses = new Class[] { Event.class, Deadline.class, Todo.class };
```
resulted in a warning due to it being an unchecked cast.

src/main/java/todoify/storage/InternalPath.java
wxwern: Thanks for highlighting this. I think I would rename this to `getWithAppendedLastComponent` instead, and similarly for other methods, to comply with our conventions.

---

The reason this was named in continuous tense was actually based on a different naming convention to imply that the "action" is not performed on the instance in question, but rather returns a new instance, such as the following:
- [Apple's Swift: `URL#appendPathComponent`](https://developer.apple.com/documentation/foundation/url/1780397-appendpathcomponent), which updates the instance in place
- [Apple's Swift: `URL#appendingPathComponent`](https://developer.apple.com/documentation/foundation/nsurl/1410614-appendingpathcomponent), which returns a new instance

This would reduce confusion as to whether an action is self-mutating or returns a new copy, since present tense often implies the former in a number of languages.

src/main/java/todoify/taskmanager/task/Deadline.java
wxwern: Do you have any suggestions for this? Since this just a string formatter, with appropriate null checks to catch unexpected errors when feeding it into a utility function, I have no idea as to how this can be further simplified.

src/main/java/todoify/storage/InternalStorage.java
wxwern: Agreed that this is confusing to new readers (it was actually errorneously left out during a refactor). I'll make it more descriptive and include some implementation details as docs.

PR #69, opened 2023-08-26
[LoMaply] iP

LoMaply:
# 🪶 Jyuuni 🪶 
> "Easy, Fast, Efficient." -Me
> "Truly one of the chatbots of all time." -Myself
> "For free?" -I
## 💬 Description
Jyuuni is a helpful chatbot aiming to help users organise their tasks. Some of its more useful features include:
- **Todo/Deadline/Event** 3 preformatted task types users can add to their list.
- **Mark/Unmark** tasks as completed or uncompleted.
- **Search** for specific tasks using keywords.

***And more*** all in an easy to use command line interface.

## ▶️ Running the project
1. [Download/Fork](https://github.com/nus-cs2103-AY2324S1/ip) the project.
2. Open it in an IDE of choice.
3. Navigate to **src/main/java/gui/Launcher.java**.
4. Run the following main method.
```java
    Application.launch(Main.class, args);
}
```
### 📔 Additional Note
To change the default location of the data file used to load and store chatbot data, navigate to 
**src/main/java/duke/Duke.java** and change the file location used in the storage field of the Duke constructor.
```java
public Duke() {
    storage = new Storage("./data");
    taskList = new TaskList();
}
```

## ⌛ ~~Future~~ Tentative Features & Updates
- [x] Add/Delete/Mark/Unmark tasks.
- [x] List/Search tasks.
- [x] Load/Save data across sessions.
- [x] An all new GUI for easier user access.
- [ ] View schedule for a chosen day???

Could [this](https://www.youtube.com/watch?v=E2xNlzsnPCQ&t=111s) be butterfly?


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+     * Parses and acts on user input for chatbot.
+     *
+     * @param userInput String to be parsed.
+     * @param taskList TaskList to be modified based on command used.
+     * @return boolean to determine if chatbot should wait for next command.
+     */
+    public static boolean parseCommand(String userInput, TaskList taskList) {
+        String[] splitInput = userInput.split(" ", 2); // Limit 2 to only separate out command word.
+        switch (splitInput[0]) {
+        case "mark":
```
andytoh1, line 194: I like that the coding standard of not indenting case clauses is adhered to.

src/main/java/taskutil/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    public String toString() {
+        return String.format("[%s] %s", this.getStatus(), title);
+    }
+
+    /**
+     * Returns a base string to be used for storing in data file.
+     *
+     * @return Formatted string, to be further modified by overriding methods in child classes.
+     */
+    public String toFileString() {
```
andytoh1, line 77: Perhaps naming these methods "getString" and "getFileString" would better adhere to the requirement of method names being verbs.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+            break;
+        case "list":
+            Ui.output(taskList.listToString());
+            break;
+        case "end":
+            Ui.output("Come back if you need anything else!");
+            return false;
+        default:
+            Ui.output("Sorry, I don't recognise this comment :(");
+            break;
```
andytoh1: Perhaps this break statement can be omitted, as was done in the Storage class.

andytoh1 (commented):
Very clean code with stringent and graceful error handling.

src/main/java/taskutil/Task.java
LoMaply: Personally I feel that using **getString** and **getFileString** might suggest that the Task object has a String variable that can be "obtained" (which to be fair it has a title variable but it isn't meant to be accessed directly).

Thus using **toString** and **toFileString** keeps it more in line with Java's default **toString** method for objects where we are simply converting an object to its string representation.

But thanks for the suggestion anyway!

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ParserTest {
+
+
+    @Test
+    public void parseDate_exceptionThrown_success() {
+        try {
+            Parser.parseDate("idk");
+        } catch (DateTimeParseException e) {
```
ruiyangzh, line 15: To make this test more complete you can have a fail() after line 14, to represent a test fail if the code executes without the correct exception. You can also use assertThrows instead of a try/catch.

src/test/java/taskutil/TaskListTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        Assertions.assertEquals("D |   | Finish homework | 12.04.2020 1530\n", tasks.listToStringData());
+
+        tasks.deleteTask(0);
+
+        Assertions.assertEquals("", tasks.listToStringData());
+    }
+
+    @Test
+    public void changeStatus_listToStringData_correctOutput() {
```
ruiyangzh, line 30: Nice compliant method names on the Junit tests.

src/main/java/taskutil/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getStatus() {
+        return (isDone ? "X" : " ");
+    }
+
+    /**
+     * Modifies status of current task.
+     *
+     * @param isCompleted New status of task.
+     */
+    public void changeStatus(boolean isCompleted) {
```
ruiyangzh: maybe a bit of a nitpick, but should it be setStatus? (since it could be the same as the current status, e.g)

src/main/java/taskutil/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Gets completed status of task.
+     *
+     * @return X if task is completed, a blank space otherwise.
+     */
+    public String getStatus() {
+        return (isDone ? "X" : " ");
+    }
```
ruiyangzh, line 42: getStatus seems to imply the method is returning a boolean, not a String.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+            } catch (DateTimeParseException e) { // Date not formatted properly
+                Ui.setOutMessage("Try the date format [dd.mm.yyyy tttt]:"
+                        + "\n       eg. [05.08.2020 1500] for 5 Aug 2020, 3PM");
+            }
+            break;
+        case "event":
+            // Using 1 split statement that splits by /(from|to) allows wrong combos like "/to x /from x".
+            try {
+                String[] eventVar = splitInput[1].split(" /from ", 2);
+                String[] times = eventVar[1].split(" /to ", 2);
```
ruiyangzh: Another nitpick, should this be in plurals (eventVars)?

ruiyangzh (commented):
Generally not too much to say, well-thought out. more junit test types can be used to test more efficiently and extensively.

PR #72, opened 2023-08-26
[ldinghan] iP

ldinghan:
# Ding, Duke's assistant
> "He's a little confused, but he got the spirit." - Will Smith ([source](https://knowyourmeme.com/memes/he-a-little-confused-but-he-got-the-spirit))

Ding cleans up after your mess. It's
* Quirky
* Text-based
* Easy to learn
* ~FAST~ _SUPER_ FAST to use

All you need to do is,

1. Download it from [here](https://github.com/ldinghan/ip/releases/download/A-Jar/duke.jar).
1. Double-click it.
1. Add your tasks.
1. Let it manage your tasks for you 🤯 

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
+package duke;
+
+import duke.commands.Command;
+import duke.exceptions.*;
```
EricXiong420: Should this be split into multiple imports (instead of wildcard)? Wildcard imports aren't allowed in the course, and will be flagged out by CheckStyle

src/main/java/duke/exceptions/InvalidDateTimeException.java
```diff
@@ -0,0 +1,16 @@
...
+public class InvalidDateTimeException extends Exception {
+    /**
+     * Constructs an duke.exceptions.InvalidDateTimeException with the specified detail message.
+     *
+     * @param message The detail message explaining the cause of the exception.
+     */
+    public InvalidDateTimeException(String message) {
+        super(message);
+    }
+}
```
EricXiong420, line 16: Missing new line after the last line. This is present in a couple of other files too. Will be flagged up by CheckStyle.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,164 @@
+package duke.parser;
+
+import duke.commands.Command;
+import duke.exceptions.*;
+import duke.tasks.*;
```
EricXiong420: Similar to previous: shouldn't be using wildcard imports. Consider importing only as needed!

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,141 @@
...
+    private TaskList setupTasks() {
+        TaskList tasks = new TaskList();
+        try {
+            File directory = new File(directoryPath);
+            if (!directory.exists()) {
+                directory.mkdir();
+            }
+            File taskFile = new File(filePath);
+            if (taskFile.createNewFile()) {
+                //new file created
```
EricXiong420: Consider adding a space after the "//", may appear neater.

EricXiong420 (commented):
Overall, I found your code easy to read for the most part except a few places where the concatenation of strings was a little confusing. Consider using something like String.format() to make those sections more readable. I noted a few minor coding standard violations too. The splitting into classes was well done!

src/main/java/duke/parser/Parser.java
pzl111: Agreed, but I think there might be some careless autocorrecting by IntelliJ? On my version of IntelliJ, auto importing all the classes within a package will automatically collapse the import statement to use *?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,141 @@
...
+                    break;
+                case 'E':
+                    description = description.split(" \\(from: ")[0];
+                    String from = String.join("", str.split("\\(from: ")[1]).split(" to: ")[0];
+                    String to = str.split(" to: ")[1];
+                    LocalDate fromDate = LocalDate.parse(from.split(" ")[0]);
+                    LocalTime fromTime = LocalTime.parse(from.split(" ")[1]);
+                    LocalDate toDate = LocalDate.parse(to.split(" ")[0]);
+                    LocalTime toTime = LocalTime.parse(to.split(" ")[1].replaceAll(".$", ""));
+                    Event taskE = new Event(description, fromDate, fromTime, toDate, toTime);
```
pzl111: Maybe you could consider extracting out the functionalities within each case into its separate functions? So the switch block will not be so long as it could be hard to trace the different cases when they are so many lines apart?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
...
+                case DELETE:
+                    Ui.deleteTaskReply(Parser.taskToDelete(str, tasks), tasks);
+                    break;
+                case FIND:
+                    Ui.findTaskReply(str, Parser.findKeyword(str, tasks));
+                    break;
+                case BYE:
+                    isRunning = false;
+                    Ui.exitReply();
+                    return;
```
pzl111: Perhaps you could consider handling the different cases separately rather than using waterfall? Might be hard to trace the flow of the program, even though waterfall will reduce code duplication

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+                String[] datetime = by.split(" ");
+                try {
+                    LocalDate date = LocalDate.parse(datetime[0]);
+                    LocalTime time = LocalTime.parse(datetime[1]);
+                    Deadline deadline = new Deadline(description, date, time);
+                    return deadline;
+                } catch (DateTimeParseException e) {
+                    throw new InvalidDateTimeException("Invalid Datetime.");
+                }
+            } else {
```
pzl111: Perhaps you could look into making the Happy Path prominent? swap out the if statement to test for the opposite, so that the unusual case is nested instead

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    @Test
+    public void testStringToTask() {
+        try {
+            Task deadline = Parser.parseStringToTask("deadline Assignment 0 /by 2023-09-04 23:59");
+            assertEquals(new Deadline("Assignment 0", LocalDate.parse("2023-09-04"),
+                            LocalTime.parse("23:59")).toString(), deadline.toString());
+        } catch (InvalidDescriptionException e) {
+        } catch (InvalidDateTimeException e) {
+        }
```
pzl111: I don't think there should be any empty catch blocks used? if so, at least provide some explanation for others to understand?

pzl111 (commented):
Overall good work, just some places could have used the Happy Path approach

PR #73, opened 2023-08-26
[PetrichorPrecipice] iP

PetrichorPrecipice:
# **Alpha**

Alpha is a Chatbot I made with my very limited programming skills. It currently is: 

- Purely text-based
- Probably disappointed at how poorly I made it
- ~~Functional~~ Usable

Here's how I developed Alpha:
1. Bid for CS2103T
2. Appeal for CS2103T
3. Spend like a solid 10 hours weekly constantly changing the code
4. And it is finally **done**. (Until next week :sob:)

Features include:
- [x] These fancy checkbox thingies that you cannot click on. 

You can go [here](https://github.com/nus-cs2103-AY2324S1/ip) to view Alpha's predecessor (even though Alpha should really be the first). 

The code I am proudest of in Alpha:
```
private static String TODO = "todo";
```

And, last but not least:

> "Bye." - Chatbot Alpha


src/main/java/alpha/FileHandler.java
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Checks or unchecks a task in the file depending on the second boolean parameter.
+     *
+     * @param index The index of the task to be marked or unmarked.
+     * @param check Boolean of whether the task is to be checked or unchecked.
+     */
+    // Checks or unchecks a task in the file. Does so by creating a temp file that copies everything over
+    // except the task that is being checked/unchecked.
+    public void checkOrUncheck(int index, boolean check) {
```
DeltonCheng: Any reason why you used a temp file instead of just overwriting the current file? Not sure if this will affect the efficiency

src/main/java/alpha/FileHandler.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+
+    /**
+     * Deletes a task from the file.
+     *
+     * @param index Index of the task to be deleted.
+     */
+    // Deletes a task by creating a temp file and copying everything but the deleted task over.
+    public void delete(int index) {
```
DeltonCheng: same as above regarding the temp file, and perhaps should use a method name that is more specific

src/main/java/alpha/Parser.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /**
+     * Returns a Command that deletes a task from storage based on an input.
+     *
+     * @return a Command that deletes a task.
+     * @throws MissingIndexException if the index is missing.
+     * @throws InvalidIndexException if the index exceeds or is lesser than the number of stored tasks.
+     */
+    public Command delete(String input) throws MissingIndexException, InvalidIndexException {
```
DeltonCheng: perhaps a more specific name for the methods, and for the other methods in several other places as well

README.md
```diff
@@ -1,24 +1,29 @@
...
-3. After that, locate the `src/main/java/Duke.java` file, right-click it, and choose `Run Duke.main()` (if the code editor is showing compile errors, try restarting the IDE). If the setup is correct, you should see something like the below as the output:
-   ```
-   Hello from
-    ____        _        
-   |  _ \ _   _| | _____ 
-   | | | | | | | |/ / _ \
-   | |_| | |_| |   <  __/
-   |____/ \__,_|_|\_\___|
-   ```
+
```
DeltonCheng, line 1: i like that you have updated the README :)

DeltonCheng (commented):
Overall, code is pretty good. Perhaps could be more specific in terms of naming the methods  and package the related files together (eg. exceptions, commands) for easier reading. Good job!

src/main/java/alpha/UI.java
```diff
@@ -42,35 +41,27 @@ public void goodbye() {
...
      * @param size Current size of the task list.
      */
-    public void taskAdded(Task task, int size) {
-        System.out.println(LINE_SEPARATOR + "\n" +
-                "Alright! I've added this task:\n " + " " + task
-                + "\nNow you have " + size + " tasks in the list.\n" +
-                LINE_SEPARATOR);
+    public String taskAdded(Task task, int size) {
+        return "Alright! I've added this task:\n " + " " + task
+                + "\nNow you have " + size + " tasks in the list.\n";
```
JasonRay168: I like how you remembered to add 8 spaces instead of 4 for these cases where the line is broken in the middle

src/main/java/alpha/UI.java
```diff
@@ -80,39 +71,29 @@ public void unmark(Task task) {
      * @param task Task that is being marked.
      * @param size Remaining size of the task list.
      */
-    public void delete(Task task, int size) {
-        System.out.println(LINE_SEPARATOR + "\n" +
-                "Noted. I've removed this task:\n" + "  " + task + "\n Now You have " + size +
-                " tasks in the list." + "\n" + LINE_SEPARATOR);
+    public String delete(Task task, int size) {
+        return "Noted. I've removed this task:\n" + "  " + task + "\n Now You have " + size +
+                " tasks in the list.";
```
JasonRay168: Maybe can move the '+' sign from the end of a line to the start of the next line? Also, make the split such that the two halves are not too different in length.

So can rewrite as something like this:
```ruby
return "Noted. I've removed this task:\n" + "  " + task
        + "\n Now You have " + size + " tasks in the list.";
```

src/main/java/alpha/Alpha.java
```diff
@@ -0,0 +1,144 @@
...
+        boolean isExit = false;
+        while (!isExit) {
+            String input = ui.read();
+            Command command = parser.parse(input);
+            command.execute();
+            isExit = command.isExit();
+        }
+        ui.goodbye();
+    }
+     */
```
JasonRay168: I think it would be better if you add a description on why you commented this code out. If there is no use for this method, then you can delete it entirely to reduce the amount of comments.

src/main/java/alpha/DeleteCommand.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class DeleteCommand extends Command {
+
+    private int index;
+
+    public DeleteCommand(TaskList taskList, FileHandler fileHandler, UI ui, int index) {
+        super(taskList, fileHandler, ui);
+        isExit = false;
+        this.index = index;
+    }
```
JasonRay168: Create JavaDoc comments for this and other classes as well, as you have done in Alpha.java and Command.java

src/main/java/alpha/Parser.java
```diff
@@ -0,0 +1,238 @@
...
+                        "\"event YOUR_DESCRIPTION " +
+                        "/from START_TIME /to END_TIME\" ");
+            }
+            return new InvalidCommand(taskList, fileHandler, ui, "");
+        } catch (DateTimeParseException e6) {
+            return new InvalidCommand(taskList, fileHandler, ui,
+                    e6.getMessage() +"The date is in an invalid format! " +
+                            "Enter the date in the format YYYY-MM-DD");
+        }
+    }
```
JasonRay168: This method is quite long, maybe instead of using if else, you can use switch case to reduce space. Alternatively can remove some of the new lines / enters in the lines so it would not consume too many lines

JasonRay168 (commented):
I think your PR code is relatively clean and easy to navigate, just a few minor coding standard issues you can work on.

PR #96, opened 2023-08-26
[inezkok] iP

inezkok:
# Beep Boop Bot

Beep Boop Bot frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is
1. download it from [here](https://github.com/inezkok/ip/releases/tag/A-Jar)
2. double-click it
3. add your tasks
4. let it manage your tasks for you :wink:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

 ```
 public class Main {
    }
}
```

### CODE REVIEW

File: src/main/java/duke/Parser.java [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Checks whether the user input for mark,unmark or delete commands is valid.
+     *
+     * @param number The string representation of the one-based index of task.
+     * @param size The size of the current task list.
+     * @throws DukeException On input error.
+     */
+    private static void validateMarkOrUnmarkorDelete(String number, int size) throws DukeException {
```
yezkez10: Maybe choose a more clear and concise name for this method, ie checkCommandValidity, or checkValidity

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    /**
+     * Check whethers the user input for creating a todo is valid.
+     *
+     * @param description The string representation of the todo description.
+     * @throws DukeException On input error.
+     */
+    private static void validateToDo(String description) throws DukeException {
```
yezkez10: Maybe use a clearer method name ie checkTodoValidity

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param list The list of tasks before being updated.
+     */
+    public void printDeleteSuccessMessage(int index, ArrayList<Task> list) {
+        appendOutput("Noted. I've removed this task:\n");
+        String taskString = String.format("%s\n", list.get(index).toString());
+        String sizeString = String.format("Now you have %d tasks in the list.\n", list.size() - 1);
+        appendOutput(taskString);
+        appendOutput(sizeString);
+    }
+}
```
yezkez10: Great abstraction of message printing!

yezkez10 (commented):
Looks good to me, some things to note:
Is the abstraction into 10 types of commands the best way? 
Is there a way to make do with lesser classes such that the overall structure is more concise and less confusing for readers?
Also, maybe you could consider changing some of the method names to be instructional so that readers can get a better idea of what each method does at first glance.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Parses through user command and make sense of it.
+     *
+     * @param input This is the user input.
+     * @param size This is the size of the current task list.
+     * @return Command This returns the respective command based on user input.
+     * @throws DukeException On input error.
+     */
+    public static Command parse(String input, int size) throws DukeException {
+        if (input.equals("bye")) {
```
CJ-Lee01: There seems to be a lot of if-else statements here. Would it be better to use a switch statement?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,198 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Checks whether the string representation of a date is valid.
+     *
+     * @param date The string representation of the date.
+     * @throws DukeException On format error.
+     */
+    private void validateDate(String date) throws DukeException {
```
CJ-Lee01: would it be better if the method returned true or false instead of throwing an exception?

CJ-Lee01: Actually nvrmind, I can see why you are throwing the duke exception.

src/main/java/duke/command/AddDeadlineCommand.java
CJ-Lee01, line 1: Looks great!
Just a thought: Would be it be good to prevent the same command object from being executed again once it has been executed?

src/main/java/duke/task/Task.java
CJ-Lee01, line 1: I think the tasks are well done!

src/test/java/duke/ParserTest.java
CJ-Lee01, line 1: I like the extensive amount of tests!

However, if you are checking to see if the function does throw an exception, junit has the `assertThrows()` method for checking the type of exceptions being thrown.

PR #97, opened 2023-08-26
[nicrandomlee] iP

nicrandomlee:
# CupidBot

> Yesterday you said tomorrow. 
> Just DO IT! ([source](https://en.wikipedia.org/wiki/Cristiano_Ronaldo))

CupidBot is your one stop task tracker! It's
- Text-based
- Fast
- _Easy to use_

All you need to do is the following
1. Download from this [link](https://github.com/nus-cs2103-AY2324S1/ip/pull/97)
2. Double-click it to run
3. Add your tasks.
4. Get reminded, all day and every day 

And it is **FREE!** 😄 

These are the things you can do with it:
- [ ] Add ToDos
- [ ] Add Deadlines
- [ ] Add Events
And **more!**

Example of Cupid Commands are:
```
todo water flowers
```
```
todo water flowers added!
```

```java
public class Main {
    }
}
```


src/main/java/functions/Load.java
```diff
@@ -0,0 +1,102 @@
...
+                    } catch (Exception e) {
+                        System.out.println("tasks.Deadline " + description + " cannot be loaded.");
+                        break;
+                    }
+                    Deadline deadline = new Deadline(description, deadlineTime, isDone);
+                    taskList.add(deadline);
+                    break;
+
+                case "E":
+                    descriptionEndIndex = currentTaskAsString.indexOf("(from:")-1;
```
Blizzeracz: Maybe you could create a new class or transfer these codes to their respective classes? 🤔 I realized the same issue in other methods in parse too!

Avoid long methods as they often contain more information than what the reader can process at a time. Take corrective action when it goes beyond 30 LoC. The bigger the haystack, the harder it is to find a needle.

src/main/java/functions/Parser.java
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * @param input file path to load
+     * @param taskList a task list containing all the tasks stored
+     */
+    public Parser(String input, TaskList taskList) {
+        this.input = input;
+        this.taskList = taskList;
+    }
+
+    public void parse() {
```
Blizzeracz: Very neat!

src/main/java/functions/Parser.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    public void deleteCommandHandler(String[] inputArray, TaskList taskList) {
+        try {
+            int targetTaskIdx = Integer.parseInt(inputArray[1]) -1;
+            Task task = taskList.get(targetTaskIdx);
+            taskList.remove(targetTaskIdx);
+            System.out.println("Noted: I've removed this task:");
+            System.out.println(task.getTaskAsString());
```
Blizzeracz: Maybe each logical units within a block should be separated by one blank line?

```suggestion
            // Get the task and task index
            int targetTaskIdx = Integer.parseInt(inputArray[1]) -1;
            Task task = taskList.get(targetTaskIdx);
            
            taskList.remove(targetTaskIdx);
            
            System.out.println("Noted: I've removed this task:");
            System.out.println(task.getTaskAsString());
```

Rationale: Enhances readability by introducing white space between logical units. Each block is often introduced by a comment.

src/main/java/functions/Parser.java
```diff
@@ -0,0 +1,223 @@
...
+                case EVENT:
+                    eventCommandHandler(taskList, input, secondHalfInput);
+                    break;
+
+                case FIND:
+                    findCommandHandler(taskList, secondHalfInput);
+
+                default:
+                    break;
+            }
```
Blizzeracz:
```suggestion
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

Note there is no indentation for case clauses.

Blizzeracz (commented):
Overall, I found your code easy to read for the most part except a few places
where the methods were too long, I think those can be abstracted out. I noted a few minor coding standard violations too. The ordering of import statement are all consistent though and the rest of the classes are quite neat. 👍

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,45 @@
+package tasks;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+
+    private String TaskIcon = "[D]";
```
Blizzeracz:
```suggestion
    private String taskIcon = "[D]";
```

 Variable names must be in camelCase.

src/main/java/functions/Save.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.StandardCopyOption;
+
+public class Save {
+    TaskList taskList;
```
Blizzeracz:
```suggestion
    private TaskList taskList;
```

Missing modifier

Blizzeracz (commented):
Adding on to the previous comment, just  small mistakes.

src/main/java/functions/Load.java
```diff
@@ -0,0 +1,102 @@
...
+            System.out.println("File does not exist in directory.");
+            return null;
+        }
+
+        Scanner s = new Scanner(f); // create a Scanner using the File as the source
+
+        while (s.hasNext()) {
+            String currentTaskAsString = s.nextLine();
+            String taskType = currentTaskAsString.substring(1,2).toUpperCase();
+            boolean isDone = currentTaskAsString.substring(4, 5).toUpperCase().equals("X");
```
xavierpok: Perhaps the arguments to .substring should be assigned variable names or have a comment explaining why those numbers were used?

src/main/java/functions/Load.java
```diff
@@ -0,0 +1,102 @@
...
+                    Deadline deadline = new Deadline(description, deadlineTime, isDone);
+                    taskList.add(deadline);
+                    break;
+
+                case "E":
+                    descriptionEndIndex = currentTaskAsString.indexOf("(from:")-1;
+                    int fromTimingStartIndex = currentTaskAsString.indexOf("(from:") + 7;
+                    int fromTimingEndIndex = currentTaskAsString.indexOf("to:")-1;
+                    int toTimingStartIndex = fromTimingEndIndex + 5;
+                    int toTimingEndIndex = currentTaskAsString.length() - 1;
```
xavierpok: Similarly, it might be difficult for new readers to understand why these numbers were used, although it might also be obvious it's just a matter of expected input.

Still, it might be good to remain clear!

src/main/java/functions/Load.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (IOException e) {
+            System.out.println("File does not exist in directory.");
+            return null;
+        }
+
+        Scanner s = new Scanner(f); // create a Scanner using the File as the source
+
+        while (s.hasNext()) {
+            String currentTaskAsString = s.nextLine();
+            String taskType = currentTaskAsString.substring(1,2).toUpperCase();
```
xavierpok: I like the clever observation that you can tell a lot of task types just from their first names, wish I had thought of that!

However, it might cause issues down the line if just strings are used and passed around. Maybe we could convert to an enum? Maybe instead of acting on the output of this string, we could have a codeblock to parse it into an enum type, similar to the one you've defined in Parser :

```java
private enum ChatFunction {
        LIST,
        MARK,
        UNMARK,
        TODO,
        DEADLINE,
        EVENT,
        DELETE,
        FIND
    }
```

And maybe a code block like so :

```
ChatFunction func;
switch (taskString) {
 
                case "T":
                    func = ChatFunction.TODO;
                    break;
```

It might help if the code needs to be refactored later on!

src/main/java/functions/Parser.java
```diff
@@ -0,0 +1,223 @@
+package functions;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import tasks.*;
```
xavierpok: Minor issue, but it might be best to avoid star imports

src/main/java/functions/Save.java
```diff
@@ -0,0 +1,64 @@
...
+    private String saveFilePath;
+    private static final String tempFilePath = "temp.txt";
```
xavierpok: Maybe this constant could be in all-caps w/ underscores to seperate words?
```suggestion
    private static final String TEMP_FILE_PATH= "temp.txt";
```

xavierpok:
Overall, LGTM, just a few issues that might be better off if looked at!

PR #131, opened 2023-08-27
[Tan Lip Wei] iP

lipwei1808:
### Duke Chatbot

> "Naming and remembering things is hard." - Me

Duke is ~~dumb~~ smart and can
- Track your tasks
- Delete tasks
- Mark them as completed
- Find tasks

#### Getting started

1. Head over [here](https://github.com/lipwei1808/ip) 
2. Clone the project
     - `git clone <link>`
3. Run it locally

And there your have it, **DUKE at your service :+1:**

#### Trying it out

- [ ] Add a task
- [ ] Mark a task
- [ ] Delete a task
- [ ] Find a task

#### For programmers

If you would like to edit the file, start from the entry point (main method) in Duke.java

```
        new Duke().run();
    }
```


src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,46 @@
+package duke.tasks;
+
+import java.time.LocalDateTime;
+
+/**
+ * A task object that tracks Deadlines.
+ */
+public class Deadline extends Task {
+    private final LocalDateTime by;
```
TeeRenJing, line 9: Perhaps capitalize constant names?

src/main/java/duke/ui/DialogBox.java
```diff
@@ -0,0 +1,83 @@
+package duke.ui;
+
+import java.io.IOException;
+import java.util.Collections;
+
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+import javafx.fxml.FXML;
```
TeeRenJing, line 8: I like how imported classes are listed explicitly!

src/main/java/duke/tasks/Tasks.java
```diff
@@ -0,0 +1,72 @@
+package duke.tasks;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a collection of Task instances.
+ */
+public class Tasks {
+    private final List<Task> tasks = new ArrayList<>();
```
TeeRenJing, line 10: I like how plural form is used on names representing a collection of objects.

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Helper method to stringify LocalDateTime objects into original text string
+     * pass by the user.
+     *
+     * @param dateTime the LocalDateTime object to be stringified
+     * @return text string of the LocalDateTime object
+     */
+    public String stringifyDate(LocalDateTime dateTime) {
```
TeeRenJing, line 68: I like how method names are verbs and written in camelCase! 👍

TeeRenJing (commented):
Looks good to me! 👍 
Just a few nits to fix.

src/main/java/duke/commands/Command.java
```diff
@@ -0,0 +1,31 @@
...
+     * @throws DukeException if the Command cannot be executed due to invalid user input
+     */
+    public abstract String execute(Tasks tasks, Ui ui, Storage storage, boolean isRestoring) throws DukeException;
+
+    /**
+     * Checks if the program should exit after the execution.
+     *
+     * @return a boolean value if the program should exit
+     */
+    public boolean isExit() {
```
lunaflight: Do you think this is an non-intuitive use of a function name?

It seems that this is only ever used to check if it is an Exit command - could using `instanceof ExitCommand` be cleaner?

src/main/java/duke/commands/FindCommand.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param target the targeted string that the user is searching for
+     */
+    public FindCommand(String target) {
+        this.target = target;
+    }
+
+    @Override
+    public String execute(Tasks tasks, Ui ui, Storage storage, boolean isRestoring) {
+        List<Task> res = new ArrayList<>();
```
lunaflight: What do you think of a more descriptive variable name, such as `foundTasks`?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Parses user text into the action type and the arguments
+     *
+     * @param text the text to be parsed
+     * @return String[] where the first index is the action type and second index is the arguments
+     */
+    protected static String[] parseText(String text) {
+        String[] words = text.trim().split(" ");
+        String[] remaining = Arrays.copyOfRange(words, 1, words.length);
```
lunaflight: What do you think of a variable name with a plural s, such as `remainingWords`?

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,136 @@
...
+ * UI of the application
+ */
+public class Ui {
+    /**
+     * Returns success message for setting marked value of a task.
+     *
+     * @param isMark boolean value whether to mark the task
+     * @param task the task to be marked
+     */
+    public String showSuccessMark(boolean isMark, Task task) {
```
lunaflight, line 17: You may want to rename the function name to something more intuitive, such as `getSuccessMarkMessage`.
`show` might give off the impression that it is a `void` function.

I noticed this in other function names in this class as well.

lunaflight (commented):
Looks good, just a few nits :)

src/main/java/duke/commands/FindCommand.java
lipwei1808: That's a good idea

src/main/java/duke/parser/Parser.java
lipwei1808: Thanks for the suggestion

src/main/java/duke/commands/Command.java
lipwei1808: That's true, I think it was added as in the future if further commands were added that could lead to the function exiting, they could easily override this function.

I agree with the naming of the function, will update it to be more intuitive

src/main/java/duke/tasks/Deadline.java
lipwei1808: I think this is not really a constant in this case as if we allow for future edits to tasks, then it wouldn't really be a constant right

PR #139, opened 2023-08-27
[zhanyang01] iP

zhanyang01:
# DukeKing

> _"It's not always that we need to do more but rather we need to focus on less_" -Nathan W.Morris, author and finance expert 
[source](https://www.tameday.com/motivational-quotes-for-work-productivity/)

Having too much work to do as a student 🤢  and worry about forgetting deadlines for submissions 💀 , fret not, as DukeKing is here to save the day and allow us to reduce our stress levels. 😀 

With DukeKing, you can focus on the things that truly matter!!!! 

### Important things to do ❗ ❗ ❗ 
1. ~~Remembering task by memory~~ 
2. Doing actual work

In addition, it is, 
* text-based
* easy to learn
* ~~FAST~~ SUPER FAST to use!!

All you need to do to use this amazing service is to,
1. download it from [here](https://github.com/zhanyang01/ip)
2. double-click on the .jar file
3. add tasks
4. let it manage tasks

and it is **FREE**!!! 👍 💯 

Currently it has, 
- [x] Managing tasks
- [ ] Managing datelines(coming soon)
- [ ] Reminders(coming soon) 

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/duke/Tasks/Task.java [hunk 1]
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Converts the date in the string format to a date in
+     * the localDate format
+     * 
+     * @param sDate is the string of the date
+     * @return a localDate object
+     */
+    public static LocalDate convertDatePlease(String sDate) {
```
laurenlim2112: Perhaps this method name should be phrased more as a verb?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+     *                 the program of the bot
+     * @return a boolean that allow the main application to end the program
+     *         when a particular command "bye" is given from the user
+     * @throws WrongMarkException      if asked to mark a marked file or unmark a
+     *                                 file
+     *                                 that is not marked
+     * @throws UnknownCommandException if asked commands that the bot do not
+     *                                 understand
+     */
+    public static boolean isExitOrContinue(String command,
```
laurenlim2112: This method might be a bit long; perhaps you could consider splitting it into 2 functions, with 1 function responsible for carrying out the command and another responsible for checking if the command is the "bye" command?

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,110 @@
...
+        System.out.println("_".repeat(40));
+    }
+
+    /**
+     * Describes what the bot say when you mark a task successfully
+     * 
+     * @param task Take in the task to be marked
+     */
+    public void markTask(Task task) {
+        String markingTask = "Nice! I've marked this task as done:";
```
laurenlim2112: Should this variable be named with a verb or a noun?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
+package duke;
+
+import java.io.IOException;
+
+import duke.exceptions.EmptyDetailsOfTaskError;
```
laurenlim2112: I like how you consistently list your imported classes explicitly. Good job!

laurenlim2112 (commented):
Overall, your code is quite clean, except for some areas with room for improvement in terms of readability. Well done for the iP so far!

src/main/java/Duke.java
```diff
@@ -2,6 +2,10 @@
 import java.util.Scanner;
 import java.util.ArrayList;
 public class Duke {
+
+    public enum Activity {
```
GohTengFong: Should enum be in all caps?

src/main/java/duke/Commands.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+/**
+ * List of all possible commands, used the idea of enums to list
+ * all the possible commands
+ */
+public enum Commands {
```
GohTengFong: Should enum be in all caps?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Scanner;
+
+import duke.exceptions.DukeException;
+import duke.tasks.TaskList;
+
+/**
+ * The main file that the application runs
+ */
+public class Duke {
+    protected UI helper;
```
GohTengFong: Perhaps a more intuitive attribute name here?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+    public static boolean isExitOrContinue(String command,
+            TaskList tasks,
+            UI helper,
+            Storage storage)
+            throws WrongMarkException,
+            UnknownCommandException {
+        String[] splittedCommand = command.split(" ");
+        String commandType = splittedCommand[0];
+        switch (commandType) {
+            case "bye":
```
GohTengFong: Should each "case" be at the same indentation level as "switch" based on coding standards?

src/main/java/duke/Tasks/Task.java
GohTengFong: I agree. I suggest removing the "Please" and phrasing it such as "convertDateFromStringToObj"?

src/main/java/duke/Exceptions/DukeException.java
```diff
@@ -0,0 +1,27 @@
+package duke.exceptions;
+
+/**
+ * Exceptions of Duke object, errors which duke is likely to encounter and what is done
+ * to catch and deal with the errors
+ */
+public class DukeException extends Throwable {
```
GohTengFong: I love how you have many different specific Duke execeptions that extend "DukeExeception".

GohTengFong (commented):
Similar to Lauren, I think your code is very clean! I love how all your files are split according to specific packages! :D

PR #141, opened 2023-08-27
[itssisi] IP

itssisi:
# Snow ☃️

Snow is here to help you remember the things you need to do. No need to worry about forgetting any tasks! 
Snow is,
- text-based
- user-friendly
- ~~SLOW~~ VERY FAST 🤩

All you need to is, 
1. download it from 
2. double-clicked
3. add your tasks
4. let it manage your tasks for you 🥳

And it is ~~NOT~~ **FREE**!

Features: 
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders (coming soon!)

Stay tuned for more cool features!

```java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,44 @@
...
+                Todo.setTodo(userOutput, inputList);
+            } else if (splitOutput[0].equals("event")) {
+                Event.setEvent(userOutput, inputList);
+            } else if (splitOutput[0].equals("delete")) {
+                inputList.delete(splitOutput);
+            } else if (splitOutput[0].equals("find")) {
+                inputList.find(splitOutput);
+            } else if (splitOutput[0].equals("mark")) {
+                inputList.mark(splitOutput);
+            } else if (splitOutput[0].equals("unmark")) {
```
Originalidk, line 31: Perhaps you could assign splitOutput[0] to a variable name

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,29 @@
...
+        horizontalLine();
+    }
+
+    public static void printBye() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+    public static void horizontalLine() {
+        System.out.println("____________________________________________________________");
+    }
```
Originalidk: Perhaps having this as a constant instead of a function would work better? Also, I could not find this being used elsewhere.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Prints out new Deadline in the task list.
+     *
+     * @param userOutput User command.
+     * @param inputList  List of tasks.
+     * @throws Exception If command is invalid.
+     */
+    public static void setDeadline(String userOutput, TaskList inputList) throws Exception {
+        try {
+            String newDes = userOutput.split("deadline")[1].split("/by")[0].strip();
```
Originalidk, line 29: Not sure if the name newDes is clear.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public void run() {
+        ui.identity();
+        Scanner userInput = new Scanner(System.in); // Create a Scanner object
+        String userOutput = userInput.nextLine(); // Read user input
+        while (!userOutput.equals("bye")) {
+            Boolean isBye = Parser.command(userOutput, inputList);
+            if (isBye) {
+                break;
+            }
```
Originalidk, line 33: Not sure if the break is necessary since your while loop should end once the user types in "bye".

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,52 @@
...
+        try {
+            String todoCommand = userOutput.split("todo")[1].strip();
+            if (todoCommand.equals("")) {
+                throw new EmptyException("todo");
+            }
+            Todo newTodo = new Todo(todoCommand);
+            inputList.add(newTodo);
+            System.out.println("Got it. I've added this task:");
+            System.out.println(newTodo);
+            System.out.println("Now you have " + inputList.size() + " tasks in the list.");
```
Originalidk, line 30: Perhaps you could create a separate method to print the statements. Noticed this elsewhere too.

Originalidk (changes requested):
LGTM. Just a few nits I noticed.

src/main/java/duke/Ui.java
raydenlim: @Originalidk its used on line 12 and 15 in Ui.java haha, 
Maybe you could change the naming to printHorizontalLine(), so that it is clearer to users what the command does :)
i.e print a horizontal line

src/main/java/duke/Ui.java
raydenlim, line 1: Solid work! Keep it up :)
Perhaps you might want to consolidate all the System.out.println(...) to specific methods in the Ui file, and update the corresponding print methods in each file with the methods. This will make the code cleaner and improve clarity.

This will also make the code easier to maintain in the long run, as you only need to edit 1 file, the ui.java file, instead of going through all the files (e.g task, event, deadline etc) to update the print statements.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,65 @@
...
+            inputList = new TaskList(storage.load());
+        } catch (Exception e) {
+            System.out.println(e);
+            ui.showLoadingError();
+            inputList = new TaskList();
+        }
+    }
+
+    public void run() {
+        ui.identity();
```
raydenlim: the name identity() is a little ambiguous for me. 
> Use nouns for things and verbs for actions

Perhaps showIdentity()?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,44 @@
+package duke;
+
+/**
+ * Parser splits the user command to be processed by the bot.
+ */
+public class Parser {
```
raydenlim, line 6: Wow, diddnt think about using parser as a boolean. This is so cool and neat!

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,28 @@
+package duke;
+
+public abstract class Task {
```
raydenlim, line 3: I think its abit weird to call this an abstract class when theres no abstract method.
Maybe my understanding is wrong haha, I'm sorry in advance

PR #150, opened 2023-08-27
[supermii2] iP

supermii2:
# Project Rock
## The simple chatbot YOU need to run keep track of your daily tasks.

Rock is a chatbot created by Alvis Ng (supermii2) for the CS2103T individual project.
It is a simple-to-use programme created in java to allow users to keep track of their tasks.

> **You should get this bot** - Me 

## Features! 👍
- Add To-Do, Deadline and Event tasks to a task list.
- Automatically save the task list and load it where you left off!
- Mark tasks as complete and remove them as necessary!
- Use tag-based commands to intuitively add tasks with the information you need
- See below for a list of commands

## How to use it
1. Download a release and place it in an empty folder.
2. Open the command window in that folder
3. Run the command `java -jar Rock.jar`

```java
public class Main {
        Application.launch(Gui.class, args);
    }
}
```

## Commands
Below is a table of commands that can be used to interact with the bot.\
All commands are sent in the format of `[command] [default input] [tagged inputs]`\
For example, to add a deadline task with name `Assignment` and deadline `2001-01-01`, type
`deadline /by 2001-01-01` \
Tagged inputs can be placed in any order!

|  Command   |                                               Description                                                |                  Default Input                  |                                                   Tagged Input                                                    |
|:----------:|:--------------------------------------------------------------------------------------------------------:|:-----------------------------------------------:|:-----------------------------------------------------------------------------------------------------------------:|
|    todo    |                          Creates a ToDo Task and inserts it into the task list                           |                  Name of Task                   |                                                       None                                                        |
|  deadline  |             Creates a Deadline Task, with a deadline time and inserts it into the task list              |                  Name of Task                   |                              &bull; /by : Date of Deadline time in YY-MM-DDDD format                              |
|   event    |         Creates an Event Task, with a start time and end time, and inserts it into the task list         |                  Name of Task                   | &bull; /from : Date of start time in YY-MM-DDDD format <br/>&bull; /to Date of Deadline time in YY-MM-DDDD format |
|    mark    |                              Marks the task at the given index as completed                              |  Index of task to mark as a numeral (e.g. "1")  |                                                       None                                                        |
|   unmark   |                             Marks the task at the given index as uncompleted                             | Index of task to unmark as a numeral (e.g. "1") |                                                       None                                                        |
|   delete   |                                     Removes task at the given index                                      | Index of task to remove as a numeral (e.g. "1") |                                                       None                                                        |
|   reset    |                                   Removes all tasks from the task list                                   |                      None                       |                                                       None                                                        |
|    list    |                                       Lists all tasks in task list                                       |                      None                       |                                                       None                                                        |
| listbydate |                              Lists all task whose by date is the one given                               |          By date in YYYY-MM-DD format.          |                                                       None                                                        |
|    find    |                         Lists all task which have the same name as the one given                         |           Name of task to search for            |                                                       None                                                        |
|    bye     | Terminates the program. Program is left open for user to see logs but no further commands can be entered |                      None                       |                                                       None                                                        |

## Acknowledgements
This project was built upon the base [Duke](https://github.com/nus-cs2103-AY2324S1/ip) provided by the CS2103T team!
Much of the GUI was heavily adapted from [se-education](https://se-education.org/guides/tutorials/javaFx.html)

Thank you to [Gracia](https://twitter.com/Hyrchurn/) and Jeremy for kindly providing me with user and bot avatars to use!


src/main/java/client/Rock.java
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+    }
+    /* Methods for Interactions with task list */
+    /**
+     * Adds the given task to
+     * the task list
+     * @param task Task to be added
+     */
+    public void addTask(Task task) {
```
marcellaantania: Since all these methods related to TaskList are not directly used in this class, would it be better to put them in the other related classes instead? For example, everything related to TaskList can be put in the TaskList class.

src/main/java/io/Parser.java
```diff
@@ -0,0 +1,51 @@
...
+    /** Dictionary of tagged inputs*/
+    private Dictionary<String, String> taggedInputs;
+    /**
+     * Creates a parsed representation
+     * of the input string given by
+     * the user's input.
+     * @param input
+     */
+    public Parser(String input) {
+        taggedInputs = new Hashtable<>();
```
marcellaantania: Is there a reason why a hashtable is needed here? Since it seems like in all instances there will only be one element inside the hashtable. Would a simple array suffice?

src/main/java/io/Invoker.java
```diff
@@ -0,0 +1,40 @@
...
+    private static String removeFirstWord(String sentence) {
+        String[] words = sentence.split(" ", 2);
+        return words.length > 1 ? words[1] : "";
+    }
+    /**
+     * Used to handle a given user input and call the corresponding method.
+     * @param inputString User's input.
+     */
+    public String handle(String inputString) throws RockException {
+        Parser input = new Parser(removeFirstWord(inputString));
```
marcellaantania: Is the first word not needed for Parser? Is it not needed for the defaultInput?

src/main/java/io/Parser.java
```diff
@@ -0,0 +1,51 @@
...
+        for (int i = 1; i < phrases.length; i++) {
+            String[] words = phrases[i].split(" ", 2);
+            taggedInputs.put(words[0], words[1]);
+        }
+    }
+    /**
+     * Get the untagged input
+     * @return Untagged input of parsed message
+     */
+    public String getDefaultString() {
```
marcellaantania: I think the naming could be more intuitive. It is not so clear what defaultString or defaultInput means. The JavaDoc could also be more descriptive.

src/main/java/io/Parser.java
```diff
@@ -0,0 +1,51 @@
...
+        return defaultInput;
+    }
+    /**
+     * Get the tagged input with the
+     * given tag
+     * @param tag Tag used to find tagged input.
+     * @return Tagged input of the given tag
+     * @throws NoSuchElementException Tag not found in the inputs.
+     */
+    public String getTaggedInput(String tag) throws NoSuchElementException {
```
marcellaantania: I was not sure what "tag" refers to in this code and had to reread multiple times to find out. Consider giving variable names that correspond to what it refers to in real life, such as taskType, commandType or similar.

src/main/java/commands/Command.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.function.Function;
+
+import client.Rock;
+import io.Parser;
+/**
+ * Representation of a command
+ * that can be issued to the chatbot.
+ * @author Alvis Ng (supermii2)
+ */
+public abstract class Command implements Function<Parser, String> {
```
marcellaantania: Is there a reason for the class Command to implement Function? Since it seems like the default apply(object) method is never used, correct me if I'm wrong.

src/main/java/client/Rock.java
```diff
@@ -0,0 +1,150 @@
+package client;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.function.Predicate;
+
+import commands.Commands;
```
marcellaantania: I like how you package everything into different packages according to their functionalities!

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Representation of a task
+ * recorded by the chatbot.
+ * @author Alvis Ng (supermii2)
+ */
+public abstract class Task implements Serializable {
+    /** Name of task */
+    protected String taskName;
+    /** One letter abbreviation for the task. */
+    protected String oneLetterAbbrev;
```
marcellaantania: This is a good idea! I like that the variable name is intuitive. The comment helps too!

marcellaantania (commented):
Your code has clear file organization which I think is great. However, there are some aspects that I think can be improved to increase the intuitiveness and readability of the code. Good job overall! :)

src/main/java/client/Rock.java
supermii2: Good catch! The methods do already exist in TaskList, but this was an attempt to decouple TaskList from Commands such that I didn't need to pass in the TaskList as an argument or create a getTaskList() method, but I think you're right in that I should just get the TaskList and use methods from there, since manipulation of the TaskList seems like a core part of the majority of commands!

src/main/java/io/Parser.java
supermii2: For Event, there are two inputs, from and to! This was my solution to preserving the input format given by the example, and essentially allows for inputs similar to kwargs in python! Not sure if it was necessary, but I think it can help futureproof in the event of a command with many different inputs.

src/main/java/io/Invoker.java
supermii2: Thank you! The first word is used to identify the command being called. The remainder of input contains the defaultInput and taggedInputs. I'll add a comment there for clarity!

src/main/java/io/Parser.java
supermii2: Thank you! I was struggling to come up with a name so came up with that. If you have any suggestions I would appreciate it! Maybe untagged or raw input.

src/main/java/commands/Command.java
supermii2: Each command is expected to return a response to the user (a response) to let the user know the task has been completed, which is why it implements a function thats takes in a parsed input and outputs a string. The apply method is used in Invoker::handle

src/main/java/io/Parser.java
supermii2: In the example given, the input for, say, an event object is given by

```deadline [Name] /from [start] /to [end]```

The tag refers to the /from and /to added and is a way for commands to extract [start] by requesting the tag "from" from the parsed input.

Admittedly it isn't an intuitive name, but because this logic is shared by all commands (to prevent some boiler plate code), I couldn't come up with a less generic name unfortunately. However, I'll try to explain the tag system more clearly in the javadocs

File: src/main/java/tasks/Task.java (hunk 8 above)
Comment (supermii2) [reply src/main/java/tasks/Task.java]: Thank you!

File: src/main/java/tasks/Task.java [hunk 9]
```diff
@@ -0,0 +1,61 @@
...
+    /** Whether the task has been completed */
+    private boolean completed;
```
dickongwd: You can consider renaming your boolean variable to `isCompleted`. Just a small nitpick!

src/main/java/commands/CommandReset.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+    /**
+     * Terminates chatbot.
+     * @param input Unused.
+     */
+    @Override
+     public String apply(Parser input) {
+        this.client.resetTaskList();
+        this.client.saveFile();
+        return ("Task List reset.");
```
dickongwd: I think the brackets around your strings can be a bit unnecessary.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * String representation of Task
+     * @return String representation of task
+     */
+    @Override
+    public String toString() {
+        return this.isCompleted()
+            ? "[" + this.oneLetterAbbrev + "][X] " + this.taskName
+            : "[" + this.oneLetterAbbrev + "][ ] " + this.taskName;
+    }
```
dickongwd: Just a suggestion to consider using `String.format` so the string's format is easier to see at first glance!

src/test/java/io/ParserTest.java
```diff
@@ -0,0 +1,18 @@
+package io;
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ParserTest{
+    @Test
+    public void defaultStringTest() {
+        String TEST = "Test String 123/by 123312/from 1394";
```
dickongwd: If you intend for your `TEST` variable to be constant, consider using the keyword `final`.

dickongwd (commented):
LGTM! Your code looks very clean and follows closely with the recommended coding standards. Just some small nitpicks and suggestions!

src/main/java/tasks/Task.java
supermii2: Thank you! Good catch

src/test/java/io/ParserTest.java
supermii2: Thank you! I missed this throughout the rest of the repo too. Good catch!

PR #157, opened 2023-08-27
[xavierpok] iP

xavierpok:
# **Kniaz**

> I've invented a new spinny thing! I think I shall call it a "weel" -Me, being unsubtle
For similar products, look [here](https://www.google.com/search?q=task+list+app&oq=task+list+app&aqs=chrome..69i57j0i512l9.6575j0j7&sourceid=chrome&ie=UTF-8&bshm=rime/1)

## New Features

- More commands
- Cleaned up design -- Each command has its own `CommandHandler`
- Extensible
- Command-Line Interface

## How to call from another java program 

```java
        KniazSession.init().run();
    }
```

## Feature map

- [x] Making tasks
- [x] Marking/unmarking tasks as done
- [x] Associating certain tasks with date/times
- [x] Deleting tasks
- [x] Searching for tasks
- [ ] GUI

### Naming explanation

1. Kniaz is a slavic equivalent of Duke, apparently
2. I thought it sounded cool 😮


src/main/java/ui/inputparser/KniazLineParser.java
```diff
@@ -0,0 +1,122 @@
+package ui.inputparser;
+
+
+import exceptions.syntax.MissingUnnamedArgsException;
+import main.logic.command.CommandFactory;
+import main.logic.command.KniazCommand;
+
+import java.util.*;
```
prawnzyy: I think try to import the exact class and avoid doing star import so as to make it clearer what class you are using?

src/main/java/ui/inputparser/InstructionType.java
```diff
@@ -0,0 +1,64 @@
...
+public enum InstructionType {
+    TODO("todo", 1, 0, new String[]{}),
+    DEADLINE("deadline", 1, 1, new String[]{"by"}),
+    EVENT("event", 1, 2, new String[]{"from","to"}),
+    MARK("mark", 1, 0, new String[]{}),
+    UNMARK("unmark", 1, 0, new String[]{}),
+    LIST("list", 0, 0, new String[]{}),
+    QUIT("bye", 0 , 0, new String[]{}),
+    DELETE("delete", 1, 0, new String[]{}),
+    FIND("find",1,0,new String[]{}),
```
prawnzyy: Try to leave a space after a comma just like your above statements

src/main/java/main/logic/command/CommandFactory.java
```diff
@@ -0,0 +1,54 @@
+package main.logic.command;
+
+
+import main.logic.handler.*;
```
prawnzyy: Try to import the direct class/classes rather than using star import

prawnzyy (commented):
Overall, I found your code to be well split and the classes are well defined. There are quite a few chunks of code commented out. If they are no longer in use, try to remove them to improve readability. Also take note of your spacing and line breaks as they are inconsistent across each file. I pointed out a few small nits but other than that, LGTM!

src/main/java/main/logic/command/CommandFactory.java
xavierpok: Right, missed this, thanks!

src/main/java/ui/inputparser/InstructionType.java
xavierpok: Was a new addition, didn't notice the variance in style, will fix!

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,68 @@
...
+import task.Task;
+
+
+
+
+/**
+ * Class encapsulating a list of Tasks, supports easy string representation, refer to Task for more details about
+ * what Tasks are.
+ */
+public class TaskList extends ArrayList<Task> implements Serializable {
```
jacksonyuan256, line 15: Interesting use of inheritance here.

src/main/java/main/logic/command/CommandFactory.java
```diff
@@ -0,0 +1,62 @@
...
+                    InstructionType.DEADLINE, new DeadlineHandler(),
+                    InstructionType.EVENT, new EventHandler(),
+                    InstructionType.MARK, new MarkHandler(),
+                    InstructionType.UNMARK, new UnmarkHandler(),
+                    InstructionType.LIST, new ListHandler(),
+                    InstructionType.QUIT, new QuitHandler(),
+                    InstructionType.DELETE, new DeleteHandler(),
+                    InstructionType.FIND, new FindHandler(),
+                    InstructionType.INVALID, new InvalidHandler()
+            ));
```
jacksonyuan256: I can see a lot of effort put into building a more complex command structure, as well as the mapping of instructions to the abstracted classes. Good job!

src/main/java/storage/save/KniazLoader.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (ClassCastException e) {
+            throw new KniazRuntimeException("Could not cast savefile class, probably corrupted",
+                    "I could not interpret your savefile data. We shall start from scratch.",
+                    e);
+
+        } catch (IOException e) {
+            throw new KniazRuntimeException("Savefile does not exist",
+                    "There is no saved data of your tasklist. We shall start from scratch.",
+                    new IOException(String.format("%s does not exist", this.saveFile.getAbsolutePath())));
+            // should not normally be thrown because we verify up above, but to satisfy compiler requirements
```
jacksonyuan256, line 88: Yes I was slightly annoyed by this too

src/main/java/storage/save/KniazLoader.java
```diff
@@ -0,0 +1,95 @@
...
+        this(DEFAULT_PATH);
+    }
+
+    /**
+     * Attempts to load saved TaskList from the location that was previously provided to this class,
+     * during construction
+     * @return the loaded TaskList
+     * @throws  KniazRuntimeException when something goes wrong in retrieving data, will wrap around the root exception.
+     */
+    public TaskList load() throws KniazRuntimeException {
```
jacksonyuan256, line 48: I find it interesting that your architecture involves TaskList being an ArrayList itself and the use of streams itself, which further increases the complexity.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    public String toPrintString() {
+        String statusIcon = "";
+        if (this.isDone) {
+            statusIcon = "X";
+        } else if (!this.isDone) {
+            statusIcon = " ";
+        }
+        return String.format("[%s] %s", statusIcon, this.taskName);
+        // return in format [statusIcon] taskname
+    }
```
jacksonyuan256: 1 small nitpick is that the output can be dictated by modifying the object toString method, and then you can just return the object itself. But otherwise this is also alright!

jacksonyuan256 (commented):
Overall very well-done, complex architecture and strictly adhering to OOP. A lot of effort was put in, good job!

PR #165, opened 2023-08-27
[lunaroddity] iP

lunaroddity:
# Bareum
4780571b-72e3-41e1-b22d-eee21f3eb4b9
> _If you can't tell how near or far it is, why don't you start with the things that are within reach?
-Doeun_

🕙💯 Always organised and on top of things, Bareum **never** forgets a task! Using a text-based interface, it's
- light
- easy to learn
- easy to use!

Try out Bareum for **FREE** [here](https://github.com/lunaroddity/ip)! Just
1. Download it
2. Run it
3. Add your tasks!

## Features
- [x] Stay on top of things by letting Bareum manage your tasks, deadlines and events!
- [ ] Find a text-based interface hard-to-read? Use the Graphical User Interface (GUI) version! (coming soon)

### Like Java?
Bareum does too! ~~of the drinkable variety~~
Here's the main method!
```java
        new Bareum("./data/storedTasks.txt").run();
}
```


src/main/java/bareum/Parser.java
```diff
@@ -0,0 +1,48 @@
...
+import bareum.commands.DeleteCommand;
+import bareum.commands.IncorrectCommand;
+import bareum.commands.ListCommand;
+import bareum.commands.MarkCommand;
+import bareum.commands.UnmarkCommand;
+
+public class Parser {
+    public Parser() {
+
+    }
```
lipwei1808, line 22: Maybe you forgot to remove this?

src/main/java/bareum/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+        if (!storedTasks.exists()) {
+            try {
+                storedTasks.createNewFile();
+            } catch (IOException e) {
+                System.out.println(e.getMessage());
+            }
+        }
+        this.storedTasks = storedTasks;
+        StringBuilder savedTasks = new StringBuilder();
+        int currChar;
```
lipwei1808: Maybe the naming could be better? Since it's an integer

src/main/java/bareum/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            FileReader fr = new FileReader(storedTasks);
+            while ((currChar = fr.read()) != -1) {
+                savedTasks.append((char) currChar);
+            }
+
+            String[] allTaskInputs = savedTasks.toString(). split("\n");
```
lipwei1808, line 54: There is an extra space before the split method

src/main/java/bareum/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import bareum.Ui;
+
+public class AddDeadlineCommand extends Command {
+    String allDetails;
+
+
+    public AddDeadlineCommand(String[] commandInputs) throws BareumException {
+        if (commandInputs.length <= 1) {
+            throw new BareumException("You are missing some arguments :(\n" +
+                    "Current number of arguments: " + commandInputs.length +
```
lipwei1808: Should leave a new line after every "+"

src/main/java/bareum/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void execute(Ui ui, Storage storage, TaskList taskList) throws BareumException {
+        try {
+            DeadlineTask task = DeadlineTask.makeDeadline(allDetails);
+            taskList.addTask(task);
+            storage.saveNewTask(task);
+            String added = "I have added this task:\n" + task + "\nYou now have "
+                    + taskList.size() + " task(s) in your list.";
+            Ui.reply(added);
```
lipwei1808: Same as above

src/main/java/bareum/DeadlineTask.java
```diff
@@ -0,0 +1,61 @@
...
+        super(isDone, description);
+        this.dueDate = dueDate;
+        this.isDone = false;
+        // add exceptions for time
+    }
+
+    public static DeadlineTask makeDeadline(String allDetails) throws BareumException {
+        if (allDetails.length() == 0) {
+            throw new BareumException("Deadline description cannot be empty.");
+
```
lipwei1808: You got an extra line here

src/main/java/bareum/DeadlineTask.java
```diff
@@ -0,0 +1,61 @@
...
+import java.time.format.DateTimeParseException;
+
+public class DeadlineTask extends Task {
+    LocalDate dueDate;
+
+    private DeadlineTask(boolean isDone, String description, LocalDate dueDate) {
+        super(isDone, description);
+        this.dueDate = dueDate;
+        this.isDone = false;
+        // add exceptions for time
```
lipwei1808: I think this is for a future addition right or is this misplaced

lunaroddity: Yep, I planned to add this in a future commit and left this as a reminder to myself :") Thanks for the feedback!

src/main/java/bareum/Parser.java
lunaroddity: Wasn't really sure if I needed a constructor if it was empty so I created one just in case but in retrospect, it's redundant so I'll remove it

src/main/java/bareum/Storage.java
lunaroddity: Thanks for the feedback! I'll probably change it to something like `int currCharAsciiCode`

src/test/java/bareum/ParserTest.java
ryanongwx, line 1: Great. I feel that your test have encompass most of the scenarios that your chatbot will be put through

src/main/java/bareum/EventTask.java
ryanongwx, line 1: Your functions are well commented such that readers will immediately be able to understand your program

src/main/java/bareum/Ui.java
ryanongwx, line 1: There is a redundant space which changes the indent to 5 spaces instead of the conventional 4 spaces

src/main/java/bareum/commands/AddTodoCommand.java
ryanongwx, line 1: Good job keeping fields private

PR #169, opened 2023-08-27
[Miguel Tarcena] iP

proto-aiken-13:
#**Introducing LatiBot!**

> "I've seen things, you people wouldn't believe, hmmm.
... attack ships on fire off the shoulder of Orion.
I've watched see Beams glitter in the dark near the Tannhauser Gate.
All those moments, will be lost in time like tears in rain..."
["... time to die ..."]

Do you feel lonely? Do you hate social interaction too? Do you want to have an unjudging, supportive friend who helps you remind of tasks that you need to do?

Well, too bad. Best I can give is LatiBot.

It's,
- Text-based.
- Easy to learn
- Reads like that one high school friend who's really into anime.
- ~~Fast~~ ~~Super Fast~~ **SLIGHTLY FASTER THAN SUPER FAST!**

Instructions:
1. Download it from [here](https://github.com/proto-aiken-13/ip/releases/tag/A-Jar).
2. Run it in a terminal.
3.  Add your tasks.
4.  Substitute your social life with it. 😉

Features
- [x] Manages tasks.
- [ ] Manages deadlines (coming soon!)
- [ ] Reminders (coming soon, too!)

```public class Main {
        Application.launch(Main.class, args);
    }
}```

### CODE REVIEW

**Comment by joel-foo** [comment]
Hi, I've been assigned to review your PR, but I think in your latest commit, you removed all your src files.

**Comment by andrefoo** [comment]
> Do note! Your `UI` class could be renamed to `Ui` to be in compliance with the module's Java coding standard. Abbreviations and acronyms should not be uppercase when used as a part of a name.

File: src/main/java/duke/Task.java [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+import java.io.Serializable;
+
+/**
+ * Tasks are the tasks sent by the user to the Duke class
+ */
+public abstract class Task implements Serializable {
+    /**The name of the task.*/
+    private String taskName;
+    /**Status on whether the task is done or not.*/
+    private boolean taskStatus;
```
vivienherq: Boolean variables should be named to sound like booleans.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
+package duke;
+
+/**
+ * The Duke class helps to manage all messages sent by the user to the chatbot accordingly.
+ */
+public class Duke {
```
vivienherq, line 7: LGTM!

src/main/java/duke/DateAndTime.java
```diff
@@ -0,0 +1,70 @@
...
+        return month + " " + day + ", " + year;
+    }
+
+    /**
+     * Formats the date and time
+     * @param date The date in YYYY-MM-DD format.
+     * @param timing The time in HH:MM format
+     * @return The formatted date and time String.
+     */
+    public String dayParse(String date, String timing) {
```
tllshan: If I'm not wrong, method names should be verbs? So perhaps this could be renamed as "parseDay" instead?

src/main/java/duke/DukeList.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Helps to store user deadline tasks and deadlines into the list.
+     *
+     * @param task takes in the task given by the user.
+     * @param deadline takes in the deadline of the deadline task.
+     */
+    public void store(String task, String deadline) {
+        Task todo = new Deadline(task, deadline);
```
tllshan: I think it could get confusing if you name the Task as todo, when it's specifically a Deadline and not a ToDo. Perhaps you could consider renaming it to deadline? Also applies to the Event in the store method below (line 74).

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Helps to parse and manage the user's inputs. "list" shows the current list, "mark / unmark"
+     * helps to mark or unmark the specific task in the list, and todo/deadline/event
+     * adds todo/deadline/event tasks respectively.
+     *
+     * @param tally takes in the input string.
+     */
+    public void messageHandler(String tally) {
```
tllshan: I think method names should be verbs, instead of nouns? Maybe this could be renamed to "handleMessage"

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,173 @@
...
+        String comd = tally.split(" ")[0];
+
+        switch (comd) {
+            case "list":
```
tllshan: If I'm not wrong, the coding standard for this course states that there should be no indentation for case clauses. So I think each case should be aligned with the switch?

src/main/java/duke/Task.java
tllshan: Yep, so perhaps this could be renamed as "isDone", or "isCompleted"

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    public static String scan() {
+        return scan.nextLine();
+    }
+
+    /**
+     * Sends a greeting message to the user.
+     */
+    public static void greet() {
+        System.out.println("____________________________________________________________\n" +
```
tllshan: The coding standard for this course states to break before an operator. So perhaps you could put the + on the next line, rather than at the end of the previous. I think doing this would make your code more readable! (Also applies to the rest of the methods in this class)

tllshan (commented):
Code looks generally well done to me, good job! Just a few small suggestions here and there to improve it further.

src/main/java/duke/Task.java
proto-aiken-13: Thank you for your feedback! I missed that out...

src/main/java/duke/Parser.java
proto-aiken-13: Hello!

Thanks for your feedback, unfortunately intelliJ keeps forcing to indent these switch-case things. Is there any way to stop it from doing so?

src/main/java/duke/DateAndTime.java
proto-aiken-13: True! Thanks for the feedback!

PR #170, opened 2023-08-28
[ChuanXinNg] ip

ChuanXinNg:
Roo helps to track your tasks so that they stay organized. It's
- text-based
- personal
- ~~Easy~~ **_SUPER_ EASY** to use

All you need to do is just
1. download it from [here](https://github.com/ChuanXinNg/ip.git)
2. open it
3. add your tasks
4. leave your tasks for Roo to manage✨

The most important thing is... it is **FREE!!!**

Features:
- [x] Manage todo tasks
- [x] Manage deadlines and events
- [ ] Find a specific event
- [ ] Filter your tasks by date

You can use Roo to practice Java too. Here is the `main` method:
```java
        Roo roo = new Roo("roo.txt");
        roo.run();
    }
```


src/main/java/roo/Roo.java
```diff
@@ -0,0 +1,152 @@
+package roo;
+
+import java.time.LocalDate;
```
zhanyang01: I like how imports are done well, imported classes are listed out explicitly, Good job!

src/main/java/roo/task/Task.java
```diff
@@ -0,0 +1,78 @@
...
+    public Task(String task, boolean isFinish) throws RooException {
+        this.isFinish = isFinish;
+        this.task = task;
+    }
+
+    /**
+     * Checks if the task is marked as done.
+     * @return true if the task is marked as done, false otherwise.
+     */
+    public boolean done() {
```
zhanyang01, line 35: This should sound like a boolean, perhaps isItFinished will be better?

src/main/java/roo/Roo.java
```diff
@@ -0,0 +1,152 @@
...
+import java.time.format.DateTimeFormatter;
+
+import roo.task.Deadline;
+import roo.task.Event;
+import roo.task.Todo;
+
+/**
+ * The main class for the Roo application, which is a tasks management program.
+ */
+public class Roo {
```
zhanyang01, line 8: Class names are all in PascalCase, variables and methods are in camelCase which is good! Keep up the good work!

zhanyang01 (commented):
I like how the code standards are met in terms of readability, Keep up the good job in the IP so far!

PR #177, opened 2023-08-28
[Low Jun Yu] iP

jylow:
### B055man :smile:

<blockquote>
"I will always find a lazy person to do the hardest tasks" - Bill Gates
</blockquote>

B055man is a friendly chatbot to help keep track of things that you need to complete. It is,

* User Friendly
* Simple to pickup
* ~$29.99~ ****FREE**** to use :exclamation:

All you need to use it,
1. Have a computer
2. Click [here](https://github.com/jylow/ip/releases/tag/A-Jar) to download (I promise its not malware) :rofl:
3. Run the program
4. Use B055man to manage your tasks

Did I mention that its ****free****?

!Features of B055man:!!
- [ ] Manage tasks
- [X] Add todos, events and deadlines
- [ ] Saves previously entered tasks 

Make sure that you have downloaded the right code:
Below is the `main` method that you should see at the bottom of the `Duke` class
```ruby
}
```


src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,166 @@
...
+     * @throws InvalidTaskNumberException Error when given task number exceeds the number of tasks in the list
+     */
+    public void markTask(int taskNumber, Ui ui) throws InvalidTaskNumberException {
+        if (taskNumber > tasklst.size()) {
+            throw new InvalidTaskNumberException("Error! Task Number given is outside range of current list size of "
+                    + tasklst.size());
+        }
+        tasklst.get(taskNumber - 1).markAsDone();
+        ui.showMarked(tasklst.get(taskNumber - 1));
+    }
```
howenc: I don't like how ui is only included in the arguments just to showMarked(Task task).
Do you think making the Ui methods static would help make this code simpler?

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public void unmarkTask(int taskNumber, Ui ui) throws InvalidTaskNumberException {
+        if (taskNumber > tasklst.size()) {
+            throw new InvalidTaskNumberException("Error! Task Number given is outside range of current list size of "
+                    + tasklst.size());
+        } else {
+            tasklst.get(taskNumber - 1).unmarkAsDone();
+            ui.showUnmarked(tasklst.get(taskNumber - 1));
+        }
+    }
```
howenc: Same here, I don't like that ui is here.
Should be a better way to do this right?

src/main/java/duke/util/TaskList.java
howenc, line 1: I like the logic in this file.
It is very smooth and flows nicely.
Each method is short and concise except for that 'parser' you have within the constructor.
What I can think of is putting it in another class just to parse the data you load from the file, but that may be too much work.
The next is that I strongly believe that if you change Ui to have static methods, you would not have to deal with needing to place it in the arguments.

src/main/java/duke/task/Task.java
howenc, line 1: Short and sweet I like it.

src/main/java/duke/task/Deadline.java
howenc, line 1: Just wondering, you are adding the format with YYYY-MM-DD HHHH, but printing the format in dd MMM yyyy?
I think it would be better to be more consistent?
But if it is a design choice then more power to you.

src/main/java/duke/task/Event.java
howenc, line 1: I see `DateTimeFormatter.ofPattern("dd MMM yyyy")` a lot, are you able to make it inheritable or referred to easily?

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Checks if the current task is an exit task.
+     * @return true since task is an exit task
+     */
+    @Override
+    public boolean isExit() {
+        return true;
+    }
+}
```
howenc, line 36: I don't like this.
If you just need to return a boolean, do you think letting execute return a boolean is good enough?

howenc (commented):
Overall it is good, I just did some nit-picking.
The logic is sound and reasonable.

PR #185, opened 2023-08-28
[howenc] iP

howenc:
# **Duke Edits for Howen's IP**
## Description of changes and what they are supposed to do
- The program is built and tested using **gradle**
- I tried to do as much OOP as I can
- I have stud test classes that are just copies of the previous versions of my code that does not have any bugs

**Description of the classes I have**

1. Commands - the state of this class determines the executed task. So the instance of Commands that is created will already possess all that it needs to execute the intended task. I only configured it to take up to 3 phase inputs, phases being strings separated by "/". This way once I put `comandsInstance.execute()` it will do the task and return a Boolean to check if it succeed.

2. Parser - the Parser class is the class that turns the string input by the user and outputs a Commands object. How it works is that each phase of a input (string separated by "/") would be evaluated if needed, there would not be a NullPointerException thrown if the phase is not there. After getting the command words according to the enums COMMAND, it will do a check on the input and create and return the Commands instance accordingly.
```Java
Parser p = new Parser(String input);
Commands c = p.parse();
c.execute();
```

3. thats all the importing things. So much for
> **Description of the classes I have**

4. Click [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ) ~~if you want to see more~~

## Logic behind design of certain classes
🤡 

- [x] a heading
- [x] a bullet list
- [x] a numbered list
- [x] a task list
- [x] an emoji
- [x] a blockquote
- [x] a hyper link
- [x] inline code
- [x] some text formatting: bold, italic, stikethrough etc.


src/main/java/task/Task.java
```diff
@@ -0,0 +1,179 @@
...
+     * Calls the constructor to create a Task.
+     *
+     * @param taskName The name of the task.
+     * @param startDayDateTime The date and time of the start of the event.
+     * @param endDayDateTime The date and time of the end of the event.
+     * @return Returns a Task object.
+     */
+    public static Task of(String taskName, LocalDateTime startDayDateTime, LocalDateTime endDayDateTime) {
+        return new Event(taskName, startDayDateTime, endDayDateTime);
+    }
```
LoMaply: These constructors seem a little redundant, you could call the child class constructors directly after putting each child class into it's own file.

src/main/java/command/Commands.java
```diff
@@ -0,0 +1,351 @@
...
+import dukeExceptions.DukeException;
+
+import java.time.LocalDateTime;
+
+public class Commands {
+
+    /**
+     * List of both primary and secondary commands.
+     */
+    public enum COMMANDS {BYE, LIST, TODO, DEADLINE, EVENT, MARK, UNMARK, DELETE, BY, FROM, TO, SORT, FIND, UNKNOWN}
```
SelwynAng: I like how you classified the commands using enums too. Good job for going the extra mile!

src/main/java/task/ListOfTask.java
```diff
@@ -0,0 +1,171 @@
...
+        try {
+            listOfTask.get(index - 1).setNotDone();
+            if (print) {
+                System.out.println(listOfTask.get(index - 1).toString());
+            }
+            Storage.save(listOfTask);
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeException("Please select from index 1 to " + listOfTask.size());
+        }
+    }
```
LoMaply: To reduce some code duplication, you might consider merging both the `mark()` and `unMark()` methods and using an additional variable to decide whether to mark or unmark a particular task.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+            FileWriter writer = new FileWriter(writeData);
+            listOfTask.forEach(x-> {
+                try {
+                    writer.write(x.write());
+                    if (x.getIsDone()) {
+                        writer.write("mark " + (listOfTask.indexOf(x) + 1) + "\n");
+                    }
+                } catch (IOException e) {
+                    System.out.println("You do not have access to write to your save file");
+                }
```
LoMaply: This try catch might be a bit redundant since it is enclosed by a try catch that catches the same exception.

src/main/java/command/Commands.java
```diff
@@ -0,0 +1,351 @@
...
+
+    private COMMANDS state;
+    private String name;
+    private int index;
+    private LocalDateTime dateTime;
+
+    /**
+     * Construct a Commands object with only the primary command.
```
SelwynAng: Perhaps verbs used in JavaDocs should be singular tense (e.g. Constructs instead of Construct).

src/main/java/task/ListOfTask.java
```diff
@@ -0,0 +1,171 @@
...
+     */
+    public void addTask(String task, LocalDateTime startDayDateTime,
+                        LocalDateTime endDayDateTime, boolean print) {
+        Task temp = Task.of(task, startDayDateTime, endDayDateTime);
+        listOfTask.add(temp);
+        if (print) {
+            System.out.println("added: " + temp);
+            Storage.save(listOfTask);
+        }
+    }
```
LoMaply: Is there a way to reduce duplication for these 3 addTask methods?

Instead of having each addTask method handle a separate task, you could have a general addTask method taking in a Task object.

This Task object would be created wherever addTask is called as opposed to being created by addTask itself.

src/main/java/command/Commands.java
```diff
@@ -0,0 +1,351 @@
...
+                    && b.index == this.index
+                    && b.name == this.name) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static class TwoCommands extends Commands {
+        private COMMANDS state2;
```
SelwynAng: Could there be a better name for this variable? A more relatable variable name would be better!

src/main/java/command/Commands.java
```diff
@@ -0,0 +1,351 @@
...
+        private String name2;
```
SelwynAng: The same goes for this. A more relatable variable name would be better!

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+import task.Task;
+import ui.Ui;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
```
SelwynAng, line 20: Perhaps adding Javadocs for describing this whole class will be better.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,179 @@
...
+        /**
+         * @inheritDoc
+         */
+        @Override
+        public String write() {
+            return "todo " + super.taskName + "\n";
+        }
+    }
+
+    private static class Deadlines extends Task {
```
SelwynAng: Perhaps extracting these private nested classes as standalone public classes would be better?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,188 @@
...
+                String command2 = this.phaseTwo();
+                String command3 = this.phaseThree();
+                if (task != null) {
+                    Parser phaseTwo = new Parser(command2);
+                    Commands c1 = phaseTwo.parse();
+                    Parser phaseThree = new Parser(command3);
+                    Commands c2 = phaseThree.parse();
+                    if (c1.checkCommand(Commands.COMMANDS.FROM) && c2.checkCommand(Commands.COMMANDS.TO)) {
+                        if (c1.compareTime(c2)) {
+                            return Commands.of(Commands.COMMANDS.EVENT, task, c1, c2);
```
LoMaply: Quite deep code indentation here, is there a way to simplify this?

SelwynAng (commented):
Overall, it was a nice implementation of your Duke project. Keep up the good work!

LoMaply (commented):
Overall, I like that you took the time to extract both the Exceptions as well as the Commands. Furthermore the testing done using JUnit appears quite in depth.

However the code itself seems quite verbose, to the point where it gets quite hard to follow the logic. There are also several instances of unnecessary duplication.

src/main/java/command/Commands.java
howenc: Roger that I'll change

src/main/java/task/ListOfTask.java
howenc: Got it

src/main/java/storage/Storage.java
howenc: Um, this part here is actually a lamda expression.
So its like a class in itself?
Its like a consumer class?
That means that I need to catch that exception, or at least thats what my IDE said.

howenc: Unless I declare that the class throws a exception?

src/main/java/task/ListOfTask.java
howenc: Yup, I made a seperate private method to do this.
I did not want to break encapsulation by allowing a class outside of the task package to touch the more specific tasks.

src/main/java/parser/Parser.java
howenc: I'll try, but what do you suggest.

PR #190, opened 2023-08-28
[weeweh] iP

weeweh:
## BanterBot ##

>"Banter is often a proof of want of intelligence." - Jean de la Bruyere [(source)](https://www.azquotes.com/quotes/topics/banter.html)

BanterBot frees your mind of having to remember things you need to do. It is:
+ text-based 
+ easy to learn
+ ~~HARD~~ FUN and FAST to use
+ humorous..? 😂

Steps to get BanterBot:
1. download it from [here](https://github.com/weeweh/ip/releases).
2. add your tasks.
3. let it manage your tasks for you 😉
4. have fun!!

Guess what: it is **FREE**!

Features include:
- [x] Create and Delete Tasks (Todos, Deadlines, and Events)
- [x] Search for Tasks
- [x] Find what Deadlines are due
- [ ] Find when Events are happening (Coming Soon..?)

``` java
public class Main {
        new BanterBot("data/tasks.txt").run();
    }
}
```


src/main/java/BanterBot.java
```diff
@@ -0,0 +1,50 @@
...
+     * Starts the ChatBot and allows the User to start interacting with it.
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                Command c = Parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
```
Elijah5399: I really like the way you structured your entire program. You used ui.readCommand to form the fullCommand string, then formed the Command using Parser.parse, and finally executed the command using command.execute, providing tasks, ui and storage as arguments. Would you like to consider giving c a more appropriate name such as "command"?

src/main/java/command/Command.java
```diff
@@ -0,0 +1,47 @@
...
+        this.index = index;
+    }
+
+    /**
+     * Returns if the Command is exiting the ChatBot.
+     * @return False
+     */
+    public boolean isExit() {
+        return false;
+    }
```
Elijah5399: I like how you named this boolean function isExit to sound like a boolean! :+1:

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import helper.Storage;
+import helper.Ui;
+import task.TaskList;
+
+/**
+ * Represents a Command that specifically deletes a Task in the TaskList.
+ */
+public class DeleteCommand extends Command {
+    /** Command the user starts with to activate the DeleteCommand. */
+    public static final String COMMAND_WORD = "delete";
```
Elijah5399: I like how you named this constant in all caps! :+1:

src/main/java/helper/Ui.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Scanner;
+
+/**
+ * Represents a User Interface that receives the User input and prints out the outputs.
+ */
+public class Ui {
+    /**
+     * A line to divide responses. */
+    public static final String DIVIDER = "____________________________________________________________";
```
Elijah5399: I noticed that you consistently use JavaDoc for methods and fields. Well done! However, do you think it would be neater to make this javadoc comment a single line instead of two lines?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+        return "[D]" + super.toString() + " (by: " + date + ")";
+    }
+
+    /**
+     * Returns a String representation of the Deadline for the .txt file.
+     * @return a String representation
+     */
+    @Override
+    public String toFileString() {
+        int bool = this.isDone ? 1 : 0;
```
Elijah5399: The naming of this variable `int bool` may be a bit misleading. Would you like to rename it to something like `int doneNumeric` instead?

LuahJunYang (commented):
Nice use of OOP for the commands, I didn't consider this approach at all. Good job naming all the methods as well as they are all quite clear

PR #204, opened 2023-08-28
[Jason Ray] iP

JasonRay168:
EchoBot

EchoBot frees your mind of having to remember things you need to do. It's,

- Text-based
- Easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. Download it
2. Double-click it
3. Add your tasks
4. Let it manage your tasks for you :relaxed:

And it's FREE! :trollface:

Features:

- [x] Managing and marking tasks
- [ ] Undo most recent change (coming soon!)

If you're a Java programmer, you can use it to practice Java too. Here's the main method:

```ruby
public class Launcher {
        Application.launch(EchoBot.class, args);
    }
}
```


src/main/java/echobot/utilities/Type.java
```diff
@@ -0,0 +1,10 @@
+package echobot.utilities;
+
+/**
+ * Enum class to declare types of tasks
+ */
+public enum Type {
```
czhiruo, line 6: I like how you have ENUMS to set predefined types

src/main/java/echobot/utilities/Ui.java
```diff
@@ -0,0 +1,245 @@
...
+import echobot.exceptions.InvalidTaskNumberException;
+import echobot.exceptions.InvalidTaskTimeException;
+import echobot.exceptions.MissingTaskDescriptionException;
+import echobot.exceptions.MissingTaskNameException;
+import echobot.exceptions.MissingTaskNumberException;
+
+/**
+ * Handles user interaction
+ */
+public class Ui {
```
czhiruo, line 17: I like how the UI class displays all the strings to be printed to the user.

src/main/java/echobot/utilities/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Handles the various cases of user inputs
+     *
+     * @param tasks TaskList object that contains the list
+     * @param input Input object that contains parsed user input
+     * @param ui Ui where the helper functions are located
+     * @return True or false signifying breaking or continuing the loop
+     */
+    public String handleInput(TaskList tasks, Input input, Ui ui) {
```
czhiruo: I like how this method is clear and concise in abstracting the input and output

src/main/java/echobot/utilities/Task.java
```diff
@@ -0,0 +1,78 @@
...
+        this.isDone = false;
+        return "You have marked this task as not done\n" + "\t" + this.convertToString();
+    }
+
+    /**
+     * Converts the task object to a String text for display
+     *
+     * @return the String representation of the task
+     */
+    public String convertToString() {
```
czhiruo, line 61: Any reason why you did it this way instead of overriding the toString method?

czhiruo (commented):
Overall, your code is very well abstracted and clear

src/main/java/echobot/utilities/Task.java
JasonRay168: Simply put, I only remembered I can overwrite toString() after about a week implementing this, and at that point I didn't want to change the method fearing that the literal string representation of the tasks might be used in the future.

Turns out It was needed to extract it into the file in the hard drive, but at this point, I don't see any difference in coding implementation. It's just that the file will become more readable that way.

Which is why I had raised a concern about the file being unreadable in the forums hahaha

I'll be sure to remember this for the tP, thanks

PR #225, opened 2023-08-28
[Mohammed Faiz] iP

Mohammed-Faizzzz:
# Duke Bartholomew Hamish Montgomery (AKA Bart)

Bart frees your mind of having to remember things you need to do. It's:

* text-based
* easy to learn
* _SUPER FAST_ to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method:

```
public class Main {
    }
}
```


src/main/java/duke/Task.java
```diff
@@ -0,0 +1,71 @@
+package duke;
+
+/**
+ * The Task class represents a generic task item with a description and completion status.
+ */
+public class Task {
+    Boolean done = false;
```
wujy28: Might be better if the boolean name sounds like a boolean (according to the Textbook), something like `isDone` :)

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    String task;
```
wujy28: I think the textbook recommends having a name that is a bit more descriptive and can explain the variable to the reader more accurately. Maybe something like `taskDescription` might be more suitable?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,71 @@
...
+     * Marks the task as done.
+     */
+    public void mark() {
+        this.done = true;
+    }
+
+    /**
+     * Marks the task as incomplete.
+     */
+    public void unMark() {
```
wujy28, line 38: This is just a nitpick but I think 'unmark' can be considered a word and so maybe the method name could have been just `unmark()` HAHA

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,71 @@
...
+        this.done = false;
+    }
+
+    /**
+     * Returns a horizontal line used for separating sections.
+     *
+     * @return The horizontal line as a string.
+     */
+    public String line() {
```
wujy28: The textbook recommends using verbs for method names, so it might be better to name this method something like `getLine()`?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,71 @@
...
+        return line;
+    }
+
+    /**
+     * Returns a formatted string representation of the task.
+     *
+     * @return The formatted string representation of the task.
+     */
+    public String getTask() {
```
wujy28, line 47: It might just be me but I feel like the method name is a bit misleading 😅, as it sounds like it will return a task object and not a string. Perhaps a name like `toFormattedString()` would be better (or anything else you can think of haha)?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,249 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * The TaskList class manages a list of tasks and provides methods to interact with and manipulate the tasks.
+ */
+public class TaskList {
+    static File taskList;
```
wujy28: I think this variable name is also a little misleading as it sounds like it's a TaskList object and not a File 😅

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * Constructs a TaskList instance with a default file path for task data.
+     */
+    public TaskList() {
+        taskList = new File("./src/main/data/tasklist.txt");
+        taskList.deleteOnExit();
+    }
+    static ArrayList<Task> tasks = new ArrayList<>();
```
wujy28: I'm not sure whether this is part of the coding standard or coding quality guidelines but it might be neater if this variable is lumped together with the other class variables at the top :)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class Ui {
+    private static final String name = "Bartholomew Hamish Montgomery";
+    static final String line = "______________________________________________________________________________________\n";
+
+    /**
+     * Displays a greeting message to the user.
+     */
+    public void greet() {
+        String greeting = line + "I extend to you my utmost felicitations, User! I am " + name + "." + "\n" + "What may I do for you?" + "\n" + line;
```
wujy28: I think the line is a little too long here (>120 characters)? You can use line wrapping :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,117 @@
...
+    public void setUserInput(String newUserInput) {
+        userInput = newUserInput;
+    }
+
+    /**
+     * Checks if the user wants to exit the application.
+     *
+     * @return True if the user input is "bye", indicating an exit command; false otherwise.
+     */
+    public boolean bye() {
```
wujy28: I feel like the method names here are a little misleading as well; even though the names are verbs but they sound more like they will perform the action of listing, marking etc. instead of returning a boolean. Maybe something like `isBye()` will be more suitable?

wujy28 (commented):
Generally well done! :)
However please pay attention to certain naming conventions and follow the guidelines provided by the Textbook (under 'Code Quality')

PR #236, opened 2023-08-28
[Catherine Liang] iP

ketweeen:
## **ValeriePro** ♡

ValeriePro frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- _FAST SUPER FAST_ to use

All you need to do is,

2. double-click it.
3. add your tasks (todo, deadline, or event).
5. And it is **FREE**!

Features:

- [ ] Managing tasks

```
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+        if (userInput.equals("list")) {
+            // If user wants to check list
+            Ui.showList(taskList);
+
+        } else if (userInput.startsWith("mark")) {
+            // If user wants to mark something
+            String[] parts = userInput.split(" ");
+            if (parts.length == 2) {
+                try {
+                    int index = Integer.parseInt(parts[1]) - 1;
```
jonasongg: Might be better to reduce the code nesting in this method? Perhaps `return;` in the first `if` to remove the `else if`

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+        } else if (userInput.startsWith("todo") || userInput.startsWith("deadline") || userInput.startsWith("event")) {
+            // User creates a task
+            Task newTask = null;
+
+            if (userInput.startsWith("todo")) {
+                // If user wants to create a ToDos task
+                if (userInput.length() < 5) {
+                    // Incorrect input format for todos
+                    Ui.showError("Incorrect input format for todo");
+                } else {
```
jonasongg: Same as above; might be better to reduce code nesting here as well?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+
+        Task task = null;
+
+        switch (type) {
+        case "T":
+            task = new ToDos(description);
+            break;
+        case "D":
+            String by = parts[3].trim();
+
```
jonasongg: I like how you separated these logically separate statements by blank lines!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+            // Parse the date and time
+            LocalDate deadlineDate = LocalDate.parse(by, DateTimeFormatter.ofPattern("MMM d yyyy"));
+
+            task = new Deadlines(description, deadlineDate);
+            break;
+        case "E":
```
jonasongg: Might be better to add a default branch to ensure all bases are covered? If control flow really should not reach the default branch, it might be better to raise an exception to detect any errors in the program!

src/main/java/duke/Deadlines.java
owenyeo, line 1: I like how clean your code looks! Keep it up.

src/main/java/duke/Deadlines.java
```diff
@@ -0,0 +1,41 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A class for Deadlines tasks.
+ */
+public class Deadlines extends Task {
```
owenyeo: Perhaps the name of this class can be named "Deadline" instead? Since an instance of "Deadlines" is referring to a singular deadline.

src/main/java/duke/Events.java
```diff
@@ -0,0 +1,41 @@
+package duke;
+
+/**
+ * A class for Events tasks.
+ */
+public class Events extends Task {
```
owenyeo: Similar to Deadlines!

src/main/java/duke/ToDos.java
```diff
@@ -0,0 +1,31 @@
+package duke;
+
+/**
+ * A class for ToDos tasks.
+ */
+public class ToDos extends Task {
```
owenyeo: Same comment as Deadlines and Events

owenyeo (commented):
Your coding style is very clean looking, and great use of the appropriate methods in parser! Your JavaDocs are also on point, with each one being very clear and in line with JavaDoc conventions. Perhaps you could think of ways to make it cleaner using Enums, or with increased abstraction?

PR #245, opened 2023-08-28
[winson8222] iP

winson8222:
# DukePro
> Something Something in the month of May.

- text-based
- easy to learn
- ~~FAST~~ _REASONABLY FAST_ to use

All you need to do is,

download it from [here](https://github.com/winson8222/ip/releases/download/A-Jar/duke.jar).
double-click it.
add your tasks.
let it manage your tasks for you 😉

And it is **FREE!**

Features:

1. Managing tasks
1. Managing deadlines (coming soon)
1. Reminders (coming soon)

This is the main Duke class in `Duke.java`
```

        new Duke(LISTPATH).run();
    }
```


src/main/java/Dates.java
```diff
@@ -5,9 +5,11 @@
 
 public class Dates {
 
-    public static boolean checkDateinput(String date) {
+    static final Pattern datepattern = Pattern.compile("^(0?[1-9]|[12][0-9]|3[01])/(0?[1-9]|1[0-2])/(\\d{4}) (\\d{4}$)");
+    static final Pattern dateStringPattern = Pattern.compile("\\d{1,2} (January|February|March|April|May|June|July|August|September|October|November|December) \\d{4} \\d{2}:\\d{2}");
```
tanshiyu1999: Good idea using regular expression to represent your dates!

src/main/java/UI.java
```diff
@@ -0,0 +1,29 @@
...
+    public static void printList(ArrayList<Task> items) {
+        System.out.println(Greeting.linebreak);
+        System.out.println("Here are the tasks in your list:");
+        for(int i = 0; i < items.size(); i++) {
+            String index = String.valueOf(i + 1);
+            System.out.println(index + ". " + items.get(i).showTaskinList());
+        }
+        System.out.println(Greeting.linebreak);
+    }
+}
```
tanshiyu1999: Might be better to handle the printing of error message with your DukeException instead of printing it out with a Ui static function.

src/main/java/Greeting.java
```diff
@@ -1,5 +1,13 @@
+/**
+ * This class encapsulates the functions used to print statements when user enters and exits the program.
+ *
+ */
 public class Greeting {
     static final String linebreak = "____________________________________________________________";
```
tanshiyu1999:
```
suggestion
static final String LINEBREAK = "______________________"
```

tanshiyu1999 (commented):
Looks good!

Overall very well defined class system. Only nit pick was your final variable.

src/main/java/duke/Commands.java
```diff
@@ -0,0 +1,86 @@
...
+            if (!sc.hasNextLine()) {
+                break;
+            }
+            try {
+                String line = sc.nextLine();
+                Parser parser = new Parser(line);
+                String command = parser.getCommand();
+                CommandType given = CommandType.valueOf(command);
+
+                switch (given) {
```
evanyan13: Maybe add implement the change in swich-case style as mentioned in the java coding standard given to us: 
**No indentation as 'case'**  - [Tweak: switch-case style](https://se-education.org/guides/tutorials/intellijCodeStyle.html#tweak-switch-case-style)

src/main/java/duke/Commands.java
```diff
@@ -0,0 +1,86 @@
...
+        boolean isRunning = true;
+        do {
+            if (!sc.hasNextLine()) {
+                break;
+            }
+            try {
+                String line = sc.nextLine();
+                Parser parser = new Parser(line);
+                String command = parser.getCommand();
+                CommandType given = CommandType.valueOf(command);
```
evanyan13: Maybe can be more specific here in terms of variable naming, instead of just `given`, can put `commandGiven`?
```suggestion
                CommandType commandGiven = CommandType.valueOf(command);
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,31 @@
+package duke;
+
+import duke.storage.Storage;
+
+/**
+ * Duke is entry point of our program
+ */
+public class Duke {
+    public static final String LISTPATH = "./data/list.txt";
```
evanyan13, line 16: I like how you extracted this out as a final variable instead of hardcoding each line with the file path

src/main/java/duke/Greeting.java
```diff
@@ -0,0 +1,28 @@
...
+import duke.ui.UI;
+
+/**
+ * This class encapsulates the functions used to print statements when user enters and exits the program.
+ *
+ */
+public class Greeting {
+
+    /**
+     * This method prints out the line when user start the program
```
evanyan13, line 16: In method header comments, the first sentence can start in the form `Returns ...`, `Sends ...`, `Adds ...`.
So I guess we need not to say `This method...` because it is already the method's header comment?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,102 @@
...
+        File file = new File(Duke.LISTPATH);
+        ArrayList<Task> items = new ArrayList<Task>();
+        int len = 0;
+        Scanner s = new Scanner(file);
+        while (s.hasNext()) {
+            String line = s.nextLine();
+            if (line.charAt(0) == 'T') {
+                String[] para = line.split(" | ");
+                String description = para[4];
+                ToDo newtask = new ToDo(description);
```
evanyan13: Maybe we can use this format for variable/instance
`newTask`
Applies to some other variables in this file as well

Actually, it we already know it is a ToDo event, why not we just do this instead:
```suggestion
                ToDo todo = new ToDo(description);
```

evanyan13 (commented):
LGTM 👍  Just that some variable naming may not be camelCase like `newtask` / `newitem`

PR #252, opened 2023-08-28
[dlathyun] iP

dlathyun:
# Kora
> English speaking Korean chatbot in Singapore!
## About
Kora is..
- text-based
- easy to learn and use
- **VERY** FAST 😎 
## Features
- [x] Add tasks with dates!
- [x] Delete tasks
- [x] Find tasks
- [ ] Look nice!
## How to use?
1. download it from [this page](https://github.com/dlathyun/ip)
2. Start using!
## Additional Information
You can also use this to practice Java programming! Here's the `main` method:
```java
public class Main {
    }
}
```
Have fun using Kora!


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,54 @@
+package duke.parser;
+
+import duke.command.*;
```
Tang-Moyan: It is a good practice that imported classes should always be listed explicitly as according to the java coding standard.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private String taskType;
+
+
+    /**
+     * Class constructor of Task.
+     * @param info String that contains details of the task.
+     */
+    public Task(String info) {
+        //num = index;
```
Tang-Moyan: Can do a final code cleaning to remove the irrelevant comments, but I guess it is ok to have them at developing stage :) .

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public String getGreetMessage() {
+        String logo = "  _   _   _\n"
+                + " / \\ / \\ / \\\n"
+                + "( 안 | 녕 )\n"
+                + " \\_/ \\_/ \\_/\n";
+        String message = "Hello, I am your chatbot Kora!\nHow can I help you today?";
+        return logo + message;
+    }
```
Tang-Moyan: I think it is possible to reduce these two functions into one. For example, when you want to call greet(), you can just call `System.out.println(getGreetMessage)` instead and then print a line if needed.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Processes user input and returns a command based on data processed.
+     * @param userInput String input from user.
+     * @return Command to be executed.
+     * @throws KoraException For invalid input.
+     */
+    public static Command parse(String userInput) throws KoraException {
+        String line = "------------------------------" + "\n";
```
Tang-Moyan: I cannot see where this variable is used... In addition, do you think we should make this "line" consistent with the "line" you declared in the Command.java? I personally think it is a good practice to make them refer to a single variable. In this way, you only need to modify at one place to change your display instead of modifying every occurrence :) .

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,54 @@
...
+            } else if (userInputArray[0].contains("deadline")) {
+                command = new DeadlineCommand(userInputArray);
+            } else if (userInputArray[0].contains("event")) {
+                command = new EventCommand(userInputArray);
+            } else if (userInputArray[0].contains("todo")) {
+                command = new ToDoCommand(userInputArray);
+            } else if (userInputArray[0].contains("delete")) {
+                command = new DeleteCommand(userInputArray);
+            } else if (userInputArray[0].contains("find")) {
+                command = new FindCommand(userInputArray);
```
Tang-Moyan: The `userInputArray[0].contains("find")` will return true as long as it has a substring "find" in it. In this case, "asdfacads**find**adfaefedcef" will also be considered as a valid command. Similar for the others. What do you think?

Tang-Moyan (commented):
Generally very good in following java coding standard which I should give comments on only, as you are my first PR review :) .

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,25 @@
...
+        return commandMessage;
+    }
+    @Override
+    public void execute(TaskList taskList, Storage storage) {
+
+        commandMessage = "Byebye. See you again!";
+    }
+
+    @Override
+    public boolean isExitYet() {
```
simbayippy: can naming convention be improved? such as just isExit()

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,29 @@
...
+import duke.task.Task;
+import duke.task.TaskList;
+
+public class MarkCommand extends Command {
+
+    private String commandMessage = "";
+    private int taskIndex;
+
+    public MarkCommand(String[] details) {
+        taskIndex = Integer.valueOf(details[0].replace("mark ", ""));
```
simbayippy, line 21: what if command is not an Integer value? such as `mark a/ mark $`

simbayippy (commented):
Overall very good code quality, could not find much issues with naming convention, and code looks clear and succinct. Maybe in the future could expand on the Exceptions, throwing specific exceptions for specific cases - but still overall really well done!

src/main/java/duke/parser/Parser.java
dlathyun: Thanks for pointing it out!

src/main/java/duke/ui/Ui.java
dlathyun: Good point haha. Thanks!

src/main/java/duke/parser/Parser.java
dlathyun: Okay thanks! I think that will be better :)

src/main/java/duke/parser/Parser.java
dlathyun: Good point.. I will take note. Thanks!

src/main/java/duke/command/ByeCommand.java
dlathyun: Okay! Will change. Thanks :)

src/main/java/duke/command/MarkCommand.java
dlathyun: Thanks for pointing out. I will take note!

PR #255, opened 2023-08-28
[Kim Hanjoo] iP

hjoneweek:
### Assistant Ari
Ari helps you with organizing your tasks. You can:

1. **Add** todos, events, and deadline
2. **Mark** tasks as done
3. **Unmark** tasks
4. **Delete** tasks

The base code is [here](https://github.com/nus-cs2103-AY2324S1/ip)

It is 

- Text-based
- Easy to learn and use

- [ ] Check if you have downloaded it! 😉

> "Organize your tasks." - Kim Hanjoo

`public class Main {`
`    public static void main(String[] args) {`
`        System.out.println("Use it")`
`    }`
`}`


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,209 @@
...
+        return lst;
+    }
+
+    /**
+     * Converts text in the String to Task objects
+     * @param text a line of text in text file
+     * @return a Task object
+     * @throws DukeException if the text is not formatted correctly.
+     */
+    public static Task textToTask(String text) throws DukeException {
```
raynertjx: Perhaps this method could be renamed to `convertTextToTask`?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,117 @@
...
+        printLine();
+
+    }
+
+    /**
+     * Prints information about the task that has been marked as done
+     * @param index task number
+     * @param tasks TaskList object
+     */
+    public void printAfterMark(int index, TaskList tasks) {
```
raynertjx: Perhaps we could use more intuitive method name like `printAfterMarkTask`? You could extend this to the other methods below.

src/test/java/duke/StorageTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+import java.util.ArrayList;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class StorageTest {
+    @Test
+    public void textToTask_test1() throws DukeException {
```
raynertjx: Perhaps your method name for your test could be what you're trying to test. It could be something like `testNormalTodo` or `testInvalidTodo`.

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,40 @@
...
+import java.time.format.DateTimeFormatter;
+
+import java.util.ArrayList;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+
+public class TaskListTest {
+    @Test
+    public void findTaskUsingKeyword_test1() throws FileNotFoundException {
```
raynertjx: Good name for your test method! I think you could use the same idea to name your other tests.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+            ui.printList(tasks);
+        } else {
+            if (command.indexOf(" ") == -1) {
+                throw new DukeException("Incorrect Format of Command");
+            }
+            String identifier = command.substring(0, command.indexOf(" "));
+            String task = command.substring(command.indexOf(" ") + 1);
+            String[] arr = task.split("[/]");
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyyMMdd HHmm");
+            DateTimeFormatter formatter2 = DateTimeFormatter.ofPattern("HHmm");
```
raynertjx: I've noticed that you keep using these 2 lines in multiple places in your code. Perhaps we could abstract this out somewhere, like in a utility/helper class?

raynertjx (commented):
LGTM! Just some nits to fix. You've kept to most of the code quality guidelines, keep it up! 😄

PR #261, opened 2023-08-28
[garylow2001] iP

garylow2001:
# Chewy, your personalized chatbot assistant

> RRRRAAAWRRRRR - Chewbacca

### Why Chewy?

- track your tasks from the command-line
- datetime aware deadlines and events for added reliability
- ~easy~ **Very easy** to use

### How to run?

1. download it [here](https://github.com/garylow2001/ip/releases/tag/A-Jar)
2. run the duke.jar
3. let Chewy do the rest for you!😉

Alternatively run `main` in Duke.java
```java
        new Duke(DATA_FILE_PATH).run();
    }
```

### Features

- [ ] Manage Todos
- [ ] Manage deadlines
- [ ] Manage Events
- [ ] Find task by keyword
- [ ] Data automatically backed up


src/main/java/Tasks/Event.java
```diff
@@ -0,0 +1,83 @@
...
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents an Event that will happen. An <code>Event</code> object
+ * has a <code>LocalDateTime</code> start and end time.
+ */
+public class Event extends Task {
+    private LocalDateTime startTime;
+    private LocalDateTime endTime;
+    public Event(String description) throws DukeException {
```
SelwynAng: Perhaps adding Javadocs for your methods will help readers be able to better understand the functionality of your methods. Do consider adding it!

src/main/java/Parser.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return task The Task object created by the line after parsing.
+     * @throws IllegalArgumentException If the line is invalid.
+     */
+    public static Task parseTaskFromFileString(String lineFromFile) throws DukeException {
+        Task task;
+        String[] lineParts = lineFromFile.split(" \\| ");
+        String taskType = lineParts[0];
+        boolean isDone = lineParts[1].equals("X");
+        String description = lineParts[2];
+        switch (taskType) {
```
SelwynAng: I think that you should not indent your case statements. According to the coding standards, case statements should be on the same level of indentation as the switch statement.

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+        // Create a scanner object to read commands entered by the user
+        Scanner scanner = new Scanner(System.in);
+
+        // Start command loop
+        while (true) {
+            // Read the next line of input
+            String userInput = scanner.nextLine();
+
+            try {
+                if (userInput.equals("bye")) {
```
SelwynAng: Perhaps try refactoring this into a switch statement which may result in a cleaner-looking code. Good to standardise whether you want to use if-else or switch statements throughout your code!

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
+import Exceptions.DukeException;
+import java.util.Scanner;
 public class Duke {
+    private final Ui _ui;
+    private TaskList _taskList;
```
SelwynAng: Perhaps you should name your variable as taskList instead of _taskList. I think it's better not to let a variable name start with a hyphen. It's good practice to use camelCase for your variable names.

src/main/java/Tasks/Event.java
```diff
@@ -0,0 +1,83 @@
...
+            this.endTime = super.parseDateTime(endTimeString);
+            if (parts.length > 1) {
+                String startTimeString = parts[1].trim();
+                // Format the string and parse it into endTime
+                this.startTime = super.parseDateTime(startTimeString);
+            }
+        }
+    }
+
+    public Event(String description, String startTimeString, String endTimeString) {
```
SelwynAng: Would be nice to add Javadocs for all of your methods to better explain the functionality of your methods!

src/main/java/Duke.java
```diff
@@ -1,10 +1,78 @@
...
+    }
+
+    /**
+     * Returns Chewy's response to the user's input.
+     * @param userInput The user's input.
+     * @return Chewy's response.
+     */
+    public String getResponse(String userInput) {
+        try {
+            if (userInput.equals("bye")) {
```
SelwynAng: Maybe it would be better to use switch statements here instead of verbose if else statements?

src/main/java/Ui.java
```diff
@@ -0,0 +1,60 @@
...
+ * Each <Code>Ui</Code> method displays a message to the user.
+ */
+public class Ui {
+
+    /**
+     * Shows the welcome message when the user first opens the app.
+     * @return The welcome message.
+     */
+    public String showWelcome() {
+        String logo = "I'm Chewy,\n" +
```
SelwynAng: Perhaps the logo string does not need a "+" operator, it can just be condensed into 1 string.

PR #262, opened 2023-08-28
[WZWren] iP

WZWren:
# TrackerBot
Ever forgot something the moment you turned around after you reminded yourself to do it?

We've got you covered! Introducing TrackerBot, the:
 -  [X] Lightweight
 -  [X] Text-Based, and
 -  [X] Easy-to-Use app!
 - [ ] ~~Only downside is, you'd need to keep it up forever...~~
 - [X] **SCRATCH THAT**, it saves your tasks when you quit the app!

Want to try it out?
 1. Go to the [Release page](https://github.com/WZWren/ip/releases/tag/A-Jar)...
 1. Download the .jar file...
 1. Place it in an empty folder of your choice...
 1. ...and run it! Easy as pie. 🥧

> “Memory is the treasury and guardian of all things.” ~ Cicero

Just leave the hard parts to us, and use the rest of that space on something you'd treasure.

## What's inside?

- Ability to add 3 types of task classification, with room to expand.
- Mark, unmark and delete these tasks.
- Save and load into the jar file folder automatically
- Portable file - bring it anywhere, use it anytime.

Not interested in a jar file? Download the source code and create an entry point:
```
public class Main {
        new TrackerBot("your custom app name here").run();
    }
}
```


src/main/java/trackerbot/task/Task.java
```diff
@@ -0,0 +1,155 @@
...
+     * Factory method to generate Tasks from the save file.
+     * <p>ofSaveString expects the save string to be split before passing into Task.</p>
+     *
+     * @param type The String representation of the Task in the save file.
+     * @param args The Arguments for each task in the save file.
+     * @return A Task corresponding to the type and args of the save string.
+     * @throws IllegalArgumentException if the save file is corrupted.
+     * @throws DateTimeParseException if the DateTime field is invalid.
+     */
+    public static Task ofSaveString(String type, String... args)
```
itsNatTan, line 69: Good checks for invalid arguments

src/test/java/trackerbot/utils/ParserTest.java
itsNatTan, line 1: Very comprehensive tests

itsNatTan (commented):
Very good JavaDocs and version tracking. No complaints 11/10

src/main/java/trackerbot/command/Command.java
```diff
@@ -0,0 +1,226 @@
...
+        private AddCommand(CommandType type, String commandFields) {
+            this.commandFields = commandFields;
+            this.type = type;
+        }
+
+        public void execute(TaskList tasks, Ui ui) throws TrackerBotException {
+            ui.showMessage(tasks.add(type, commandFields));
+        }
+
+        public boolean isExit() {
```
chuababyy: I noticed that the `@Override` notation for `execute` and `isExit` function is not included for all the child classes. It would be good to include this notation so that any errors can be caught by the compiler at compile time!

src/main/java/trackerbot/command/Command.java
```diff
@@ -0,0 +1,226 @@
...
+        CommandType parsedType = CommandType.UNKNOWN;
+        for (CommandType command: CommandType.values()) {
+            if (keyword.equals(command.getKeyword())) {
+                parsedType = command;
+                break;
+            }
+        }
+
+        Command result;
+        switch (parsedType) {
```
chuababyy: Good that you included the explicit `// Fallthrough` comment in accordance with the Java coding standard in our course

src/main/java/trackerbot/task/Task.java
```diff
@@ -0,0 +1,155 @@
...
+     * @return true if the task is successfully marked to be incomplete,
+     *         false if the task is already marked as incomplete.
+     */
+    public boolean unmarkTask() {
+        if (!isDone) {
+            return false;
+        }
+        isDone = false;
+        return true;
+    }
```
chuababyy: Given that this is returning a boolean, it would be good to name it to sound like a boolean in accordance to the Java coding standard by our course. 

```suggestion
    public boolean canMarkTask() {
        if (isDone) {
            return false;
        }
        isDone = true;
        return true;
    }

    /**
     * Flags the task to be incomplete, if able.
     * @return true if the task is successfully marked to be incomplete,
     *         false if the task is already marked as incomplete.
     */
    public boolean canUnmarkTask() {
        if (!isDone) {
            return false;
        }
        isDone = false;
        return true;
    }
```

src/main/java/trackerbot/utils/Ui.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Generates UI elements for TrackerBot.
+ *
+ * @author WZWren
+ * @version A-JavaDoc
+ */
+public class Ui {
+    /** Line separators for the console between paragraphs. **/
+    private static final String FORMAT_LINE =
```
chuababyy: Good that you named your variables in accordance to the Java coding standards given in our course! Since these variables are constants, they are named in all uppercase using underscore to separate words

src/main/java/trackerbot/command/Command.java
WZWren: Ah, I must have forgotten about this. Thanks for pointing it out!

src/main/java/trackerbot/task/Task.java
WZWren: Hmm, should this really be named this way? These methods actually mark/unmark the task directly, and the boolean flag is actually to feedback to the caller whether the task has been marked/unmarked successfully.

src/main/java/trackerbot/command/Command.java
```diff
@@ -0,0 +1,226 @@
...
+        case TODO:
```
rocketninja7, line 59: I like that your case statements have the correct indentation.

src/main/java/trackerbot/gui/DialogBox.java
```diff
@@ -0,0 +1,94 @@
...
+            fxmlLoader.setRoot(this);
+            fxmlLoader.load();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        dialog.setText(text);
+        displayPicture.setImage(img);
+
+        Circle circleClip = new Circle(
```
rocketninja7, line 53: I like that your statements here have indentation of 8 spaces and that the lines are split appropriately.

src/main/java/trackerbot/task/Task.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Check if the description of the Task contains the query String.
+     * @param searchStr The query to match any point in the description.
+     * @return true, if the description contains the searchStr, and false
+     *         otherwise.
+     */
+    public boolean doesContain(String searchStr) {
```
rocketninja7, line 145: I like how you name methods that return booleans according to the standards.

src/test/java/trackerbot/utils/TaskDateHandlerTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test file for TaskDateHandlerTest. <br>
+ * This parses user input to Date. As the parse can be variable with optional
+ * fields, we test to see if the parser handles bad input properly.
+ */
+public class TaskDateHandlerTest {
+    @Test
+    public void convertInputToDate_dayAndMonthInput_shouldParseProperly() throws DateTimeParseException {
```
rocketninja7: I like how you name your test methods.

rocketninja7 (commented):
Excellent work!

PR #272, opened 2023-08-29
[Jweewee] iP

Jweewee:
### **Wiz**

> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes/)

Wiz frees your mind from having to remember things you need to do. It's:

- Text-based
- Easy to learn
- ~FAST~ _SUPER_ FAST to use

All you need to do is:

1. Download it from [here](#download-it-from-here).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉

And it is **FREE**!

Features:

- [X] - Managing tasks
- [ ] - Managing deadlines (coming soon)
- [ ] - Reminders (coming soon)

If you are a Java programmer, you can use it ot practice Java too. Here's the `main` method:

```
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
+public class Deadline extends Task{
+    protected String by;
```
TehOPanas: I feel like a more intuitive name can be given to **by**. 
For example, **endDate** or **deadline**.

src/main/java/Duke.java
```diff
@@ -1,10 +1,245 @@
+import java.io.*;
```
TehOPanas: I think it will be clearer if you explicitly import the classes you use instead of the whole package, so we can get a grasp of what you're using from the start.

src/main/java/Duke.java
```diff
@@ -1,10 +1,245 @@
...
+        }
+    }
+
+    /**
+     * Parses and handles the user's command to perform various tasks.
+     *
+     * @param command The user's command to be processed.
+     * @throws DukeException If the command is not recognised, or error
+     */
+    public void executeCommand(String command, String filePath) throws DukeException {
```
TehOPanas: For each if else case, I feel like you can add a bit more spacing for higher readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,245 @@
...
+     * @param args Command-line arguments (unused).
+     */
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Duke wizzer = new Duke();
```
TehOPanas: Same thing, I prefer more spacing.

TehOPanas (commented):
Overall did a great job on all the naming conventions and coding standard, but can use a bit more spacing in between codes. **GREAT JOB!!!!**

src/main/java/Duke.java
```diff
@@ -1,10 +1,228 @@
...
+    public void executeCommand(String command, String filePath) throws DukeException {
+        String[] separateCommand = command.split(" ");
+        System.out.println("--------------------------");
+        System.out.println("--------------------------");
+        if (command.equals("list")) {
+            System.out.println("Task List:");
+            for (int i = 0; i < tasks.size(); i++) {
+                System.out.println(" " + (i + 1) + ". " + tasks.get(i).toString());
+            }
+        } else if (command.startsWith("mark") || command.startsWith("unmark")) {
```
jean-cq: It could be better if you put the condition as a field/variable, according to the "Code quality" under "Implementation" chapter of the textbook.
>Avoid complicated expressions

src/main/java/Duke.java
```diff
@@ -1,10 +1,228 @@
...
+        String folderPath = "./data/";
+        String filePath = "./data/duke.txt";
+        try {
+            tasks = readFile(filePath, folderPath);
+        } catch (DukeException e) {
+            System.out.println(e);
+        }
+        System.out.println("Hello from " + logo +
+                "\nWhat can I do for you?\n");
+        System.out.println("--------------------------");
```
jean-cq: Since you have quite a lot of dashed lines to print, will it be better to create a field for them to avoid complicating the code?

src/main/java/Duke.java
```diff
@@ -1,10 +1,228 @@
...
+                        tasks.get(count).markAsDone();
+                    }
+                }
+            }
+            myReader.close();
+            return tasks;
+        } catch (FileNotFoundException e) {
+            createDataFolderIfNeeded(folderPath);
+            createDataFileIfNeeded(filePath);
+            writeLine(filePath);
```
jean-cq: I like the way you use the same level of abstraction here.

src/main/java/Event.java
```diff
@@ -0,0 +1,38 @@
+/**
+ * The Event class represents a task of type "Event" inherited from the Task class.
+ * It contains a description and a time interval during which the event occurs.
+ */
+public class Event extends Task {
+    protected String start;
```
jean-cq: A better naming of the field would be dateStart?

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    public Task(String description) {
```
jean-cq: For statements of different types, will it be better to leave a blank line in between?

src/main/java/Duke.java
```diff
@@ -1,10 +1,228 @@
...
+                    throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+                }
+            } else if (command.startsWith("deadline")) {
+                try {
+                    String[] parts = command.split("/by"); //   2/12/2019 1800
+                    String description = parts[0].substring(9).trim();
+                    if (description.length() == 0) {
+                        throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
+                    }
+                    String byID = parts[1].trim();
```
jean-cq: The naming of byID seems a bit confusing to the reader, would it be better to change it to dueDate?

src/main/java/Duke.java
```diff
@@ -1,10 +1,228 @@
...
+                throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+            if (command.startsWith("delete")) {
+                System.out.println(" Now you have " + tasks.size() + " tasks in the list.");
+            } else {
+                System.out.println(" I've added this task:" + "\n" + "   " + tasks.get(tasks.size() - 1).toString()
+                        + "\n" + " Now you have " + tasks.size() + " tasks in the list.");
+            }
+            writeLine(filePath);
+        } else {
```
jean-cq: these if-else statements can be greatly improved to a clearer presentation after you do A-MoreOOP. Do not nest too many things inside a method.

jean-cq (commented):
LGTM, efforts are seen in doing abstraction, but still can do more, especially after implementing OOP. Namings can be improved.

PR #288, opened 2023-08-29
[brein62] iP

brein62:
# Bob the chatbot
The best **Duke** chatbot in Singapore, JB, *and some say Batam*

Bob the Chatbot allows you to store tasks so that you don't forget them. It's:

- Simple
- Easy to use
- Text-based
- ~Many~ Very few commands to learn!

All you need to do is to...
1. Download ```duke.jar``` from [heree](https://github.com/brein62/ip/releases/tag/A-Jar).
2. Run ```java -jar (file).jar```
3. Add your tasks.
4. **PROFIT** :money_mouth_face: :money_mouth_face: :money_mouth_face:

> "I don't know why, but this is a very nice app to use" - me

Features:

- [ ] Managing different tasks (todo, reminders, events)
- [ ] Storing and saving tasks
- [X] Mark tasks as completed!

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method to store the tasks in a file called `./data/duke.txt`:
```java
    Path path = Paths.get("data", "duke.txt");
    new Duke(path).run();
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Parses the message to be processed into a Command, then returns that Command.
+     *
+     * @param message The message to be parsed.
+     * @return The Command parsed from the given message.
+     * @throws DukeException If there is an invalid input message causing an error in parsing.
+     */
+    public static Command parse(String message) throws DukeException {
+        if (message.equals("bye")) {
```
tanveersingh10: I think this method could look a lot cleaner with the switch / case syntax. You could break it up into smaller methods as well

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,127 @@
...
+            return new TaskList();
+        }
+    }
+
+    /**
+     * Saves all tasks in the chatbot into the file in the hard disk.
+     *
+     * @param tasks The list of tasks to be saved into the file.
+     */
+    public void saveTasks(TaskList tasks) {
```
tanveersingh10, line 113: I like the variable naming. Does exactly what you'd expect.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,137 @@
...
+import duke.task.Task;
+import duke.task.TaskList;
+import duke.task.Todo;
+
+/**
+ * Represents a command to add tasks (todos, deadlines, and events) into the task list.
+ */
+public class AddCommand extends Command {
+
+    enum CommandType {
```
tanveersingh10, line 19: good use of enums! something i didnt do lol

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+        super(message);
+        this.deadline = deadline;
+    }
+
+    /**
+     * Returns a String containing information within the Deadline task, formatted to be saved.
+     *
+     * @return The deadline, formatted as a String to be saved in the save file.
+     */
+    public String toSaveFormatString() {
```
tanveersingh10, line 63: very clear method naming! not too long and not too short

tanveersingh10 (commented):
LGTM!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,65 @@
+package duke;
+
+import java.nio.file.Path;
+
+import duke.command.Command;
+import duke.gui.GuiApplication;
+import duke.task.TaskList;
+import javafx.application.Application;
```
Darren159, line 8: I think making the ordering of import statements would help enhance the readability of the code.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,127 @@
...
+        Task newTask;
+        String[] lineSeq = line.split(" \\| ");
+        try {
+            boolean isValidCompletion = (lineSeq.length < 2)
+                ? false
+                : (Objects.equals(lineSeq[1], "1") || Objects.equals(lineSeq[1], "0"));
+
+            boolean isValidTodo = (line.startsWith("T") && lineSeq.length == 3);
+            boolean isValidDeadline = (line.startsWith("D") && lineSeq.length == 4);
+            boolean isValidEvent = (line.startsWith("E") && lineSeq.length == 5);
```
Darren159, line 51: I like that your naming of booleans adheres to java coding standards!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,137 @@
...
+        case EVENT:
+            typeStr = "Event";
+            break;
+        default:
+            typeStr = "";
+            break;
+        }
+        return "Add " + typeStr;
+    }
+}
```
Darren159, line 137: I like that your switch statement follows java coding standards!

Darren159 (commented):
Overall I like that your variable and method naming follows java coding standards and there is extensive use of Javadocs. LGTM!

src/main/java/duke/Parser.java
brein62: thank you for the feedback! didn't realise that switch-case syntax can also be used for strings as well so will make the necessary changes 👍 . i agree that this parse method could be broken up into multiple methods since the ```parse``` function implementation is very long

src/main/java/duke/Duke.java
brein62: thank you for the feedback! actually `import` statements aren't exactly the most well-defined in the CS2103T coding standards, so i just followed the order recommended by checkstyle:

- static imports
- `java.*` imports, followed by
- other imports (including `duke`/`javafx`)

but will check the files again and see if there are any coding standard violations for `import`s i may have missed out! 👍

PR #301, opened 2023-08-29
[kiatkat] iP

kiatkat:
# PRTS

> Even at time's end, we will meet again.

Part chatbot, part assistant, part productivity tool, and all yours. To all the workaholics who subsist on nothing but instant noodles and your sanity, PRTS is your friend, here to help.

Taking some of that cognitive load off of you, it features
- A text-based interface, for productivity gains
- A small but robust set of functions, for easy pickup
- A distinctive personality, to keep you company

Get it now! Simply
1. Visit this [link](https://www.youtube.com/watch?v=dQw4w9WgXcQ)
2. Profit

Best of all, it's _**free**_! Your wallet won't ache for this one. 🥲

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

Here's a preview of a method, `docRest`, in the code:
```java
    public String docRest() {
        if (tasks.isEmpty()) {
            return "How'd you even sleep while standing up... well, have a good rest.";
        } else {
            return "There's still lots of work that needs to be done. " +
                    "We can't afford to have you resting.";
        }
    }
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+import duke.task.Event;
+import duke.task.Todo;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+public class Parser {
+    public static Command parse(String fullCommand) throws DukeException {
+
+        String[] splits = fullCommand.strip().split(" ", 2);
```
tanshiyu1999: Oh I like that you put the split outside, I split this thing inside which result in a lot of duplicate code.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+                    throw new DukeException("Please format your description properly.");
+                }
+
+            case "event":
+
+                if (detail.isBlank())
+                    throw new DukeException("I can't read your mind. Do add a description.");
+                try {
+                    String[] eventSplit1 = detail.split("/from", 2);
+                    String[] eventSplit2 = eventSplit1[1].split("/to", 2);
```
tanshiyu1999: I think the naming of the variables does not make it very clear exactly what you are splitting into.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,54 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Ui {
+
+    Scanner scanner = new Scanner(System.in);
+    static String softbreak = "--------------------------------------------------";
+    static String hardbreak = "——————————————————————————————————————————————————";
```
tanshiyu1999: softbreak and hardbreak to be named as 
SOFTBREAK
HARDBREAK
since they are a constant string variable.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+
+import duke.command.*;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Todo;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+public class Parser {
```
tanshiyu1999: Can consider using regex to make your parser more constomizable and less reliant on String manipulation.

tanshiyu1999 (commented):
LGTM. 

Good project structuring with abstract classes.

src/test/java/prts/ParserTest.java
```diff
@@ -0,0 +1,41 @@
...
+        try {
+            assertEquals(CommandParser.parse("lIsT").getClass(), ListCommand.class);
+        } catch (ParsingException e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void parse_listMultipleWords_reject() {
+        try {
```
MadLamprey, line 35: Test case methods follow naming convention!

src/main/java/prts/Prts.java
```diff
@@ -0,0 +1,65 @@
...
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                ui.showSoftLine(); // show the divider line ("_______")
+                Command c = CommandParser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
```
MadLamprey: Good division between the uses of each - UI, Parser and Command Class

src/main/java/prts/Prts.java
```diff
@@ -0,0 +1,65 @@
...
+ * predefined commands.
+ */
+public class Prts {
+
+    private final Storage storage;
+    private TaskList tasks;
+    private final Ui ui;
+
+    private String fileName = "prts.PRTS.txt";
+    private String[] directories = {"data"};
```
MadLamprey, line 18: Why is it being stored as an array?

src/main/java/prts/Ui.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public void showHardLine() {
+        System.out.println(hardbreak);
+    }
+
+    /**
+     * Lists the current Tasks stored in the TaskList.
+     * @param tasks The current state of the TaskList.
+     */
+    public void list(TaskList tasks) {
```
MadLamprey: Could change the name of method to a more explanatory name

src/main/java/prts/command/ParsingException.java
```diff
@@ -0,0 +1,66 @@
...
+import prts.PrtsException;
+
+/**
+ * Signals that an error has occurred when parsing the user input.
+ * This error can be of various types, and a different error message is displayed according to the
+ * input type of error. A full list of error types is listed in the enum ExceptionType.
+ */
+public class ParsingException extends PrtsException {
+
+    public enum ExceptionType {
```
MadLamprey: Good use of enums instead of creating multiple sub classes

MadLamprey (commented):
Good Job!

src/main/java/duke/Parser.java
kiatkat: Point taken! Thanks for spotting that.

src/main/java/prts/Prts.java
kiatkat: It's possible that the user might want to create the file in a nested directory somewhere, so I wanted to accept multiple directory names as inputs to enable such a functionality in the future. Though I do agree that as of now, this is indeed a bit redundant as this field is effectively final.

E.g. If `directories = {"deeply", "nested", "data", "folder"};`, I'd like to be able to create the save file in the path `user.dir/deeply/nested/data/folder/`.

PR #308, opened 2023-08-29
[nubnubyas] iP

nubnubyas:
# **Monday**

> "Technology is at its best when it brings simplicity and efficiency to our daily lives."

Monday frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- FAST SUPER FAST to use

All you need to do is,

- download the zip from [here](https://github.com/nubnubyas/ip/releases/tag/A-Jar).
- double-click it.
- add your tasks.
- let it manage your tasks for you 😉

And it is **FREE**!

Features:
- [x]  Managing tasks
- [x]  Managing deadlines (coming soon)

#
```java
public class Main {
    }
}
```


src/main/java/monday/monday/exception/MondayExceptions.java
```diff
@@ -0,0 +1,27 @@
+package monday.monday.exception;
+
+/**
+ * MondayExceptions is a custom exception class for handling
+ * exceptions related to the Monday application.
+ */
+public class MondayExceptions extends Exception {
```
HugeNoob: Since the class refers to a single instance of an exception, perhaps it should be named `MondayException` (singular) instead?

src/main/java/monday/Monday.java
```diff
@@ -0,0 +1,56 @@
...
+            try {
+                isRunning = Parser.mondayParser(userInput, taskList);
+            } catch (MondayExceptions e) {
+                System.out.println(e);
+            } catch (NumberFormatException e) {
+                System.out.println("Mark/UnMark number error. " + e.getMessage());
+            } catch (IllegalArgumentException e) {
+                System.out.println("Argument Error: " + e.getMessage());
+            } catch (IndexOutOfBoundsException e) {
+                System.out.println("Index out of Bound Error: " + e.getMessage());
```
HugeNoob: Perhaps these println statements could be done through the Ui class instead? I noticed this issue in several other places too.

src/main/java/monday/task/TaskList.java
```diff
@@ -0,0 +1,135 @@
...
+            System.out.println("Error saving tasks: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Adds a task to the task list.
+     *
+     * @param task The task to be added.
+     */
+    public void addToTask(Task task) {
```
HugeNoob: Perhaps this should be named `addToList` or `addTask` instead?

src/main/java/monday/monday/parser/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Parses the user input and performs the corresponding action.
+     *
+     * @param userInput the user input to be parsed
+     * @param taskList the TaskList object to perform actions on
+     * @return true if the application should continue running, false otherwise
+     * @throws MondayExceptions if there are errors related to the Monday application
+     * @throws IllegalArgumentException if the user input is in the wrong format
+     */
+    public static boolean mondayParser(String userInput, TaskList taskList) throws MondayExceptions {
```
HugeNoob: Perhaps you could simply rename this method to `parse` instead. Currently, you would have to call `Parser.mondayParser(...)`, which seems a bit verbose. Also, method names should sound like verbs, but it currently sounds like a noun.

HugeNoob (commented):
Code quality is great in general. Verified with checkstyle to be aligning with the course's Java standards.

Side note, is there any reason why you chose to bundle `datetime`, `exception`, `parser`, and `ui` in a `monday` package? It seems a little unintuitive to have a nested package with the exact same name as its parent. I noticed that you had to do `import package.monday.monday.etc...` in many places, which seems a little strange. Perhaps you could rename the nested package or remove it and place the folders in the same level as the rest.

src/main/java/monday/storage/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+
+import monday.task.Task;
+
+
+/**
+ * Class for handling the storage of tasks in a file.
+ */
+public class Storage {
```
H1410101: It seems intuitive to me to rename this to something more specific, eg `TaskArrayListStorage`, to reflect its intention to be a lot more specific than a general object storage (which could work, but would be implemented through a generic, perhaps).

src/main/java/monday/storage/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Saves the given list of tasks to the file.
+     *
+     * @param tasks the list of tasks to save
+     * @throws IOException if an I/O error occurs while saving the tasks
+     */
+    public void save(ArrayList<Task> tasks) throws IOException {
+        try (ObjectOutputStream output = new ObjectOutputStream(new FileOutputStream(FILEPATH))) {
+            output.writeObject(tasks);
+        }
```
H1410101:
```suggestion
        output.writeObject(tasks);
```
This might be a bit redundant
(p.s. trying the suggestion feature; feel free to ignore / close / reject)

src/main/java/monday/monday/parser/Parser.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Parses the user input and performs the corresponding action.
+     *
+     * @param userInput the user input to be parsed
+     * @param taskList the TaskList object to perform actions on
+     * @return response from chatbot
+     * @throws MondayExceptions if there are errors related to the Monday application
+     * @throws IllegalArgumentException if the user input is in the wrong format
+     */
+    public static String mondayParser(String userInput, TaskList taskList) throws MondayExceptions {
```
H1410101: like the reviewer before me has mentioned, this isn't phrased as a verb :P

I'm a lot less sure of this, but:
do you think it's a good idea to make this method throw `NumberFormattingException` as well? Your code handles it properly, but I thought it might add a bit of clarity to the types of exceptions expected from this method. 
To some extent, this ambiguity is inherent for the Exception design pattern in general, though, so feel free to disagree.

src/main/java/monday/task/TaskList.java
```diff
@@ -0,0 +1,143 @@
...
+        save();
+        return Ui.addTask(task, list.size());
+    }
+
+    /**
+     * Displays the list of tasks.
+     *
+     * @return Message with the list of all tasks.
+     */
+    public String displayList() {
```
H1410101: Rather than `display`, which I believe contradicts the purpose of this method (which is to convert to a String representation), I think this is a very natural functionality to override `toString` for.

src/main/java/monday/task/TaskList.java
```diff
@@ -0,0 +1,143 @@
...
+    public String delete(int index) {
+        if (index < 1 || index > list.size()) {
+            throw new IndexOutOfBoundsException("Task index is out of range. "
+                    + "Check the number of tasks using the 'list' command.");
+        }
+        Task taskToEdit = list.get(index - 1);
+        list.remove(index - 1);
+        save();
+        return Ui.deleteTask(taskToEdit, list.size());
+    }
```
H1410101: I'm personally not a fan of referring to `Ui` from within `TaskList`; it feels like this could be achieved without "interlocking" classes this way, perhaps by shifting this printing functionality to `Parser`.

src/main/java/monday/task/Task.java
```diff
@@ -0,0 +1,57 @@
...
+     * Marks the task as done by setting its completion status to true.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Marks the task as not done by setting its completion status to false.
+     */
+    public void unMark() {
```
H1410101, line 42: I think there's an opportunity here to mirror the phrasings of related methods.
```suggestion
    public void markAsUndone() {
```

H1410101 (approved):
LGTM
Minor nits wrt naming (from previous reviewer as well), and opinionated comments/suggestions from me regarding specific code design choices, but code is quite clean. I like it!

src/main/java/monday/Monday.java
nubnubyas: That's a good suggestion, I will look into it.

src/main/java/monday/task/TaskList.java
nubnubyas: Good catch, did not realise the bad naming.

src/main/java/monday/monday/parser/Parser.java
nubnubyas: I agree with you, I'll fix that, thanks for pointing out.

src/main/java/monday/monday/parser/Parser.java
nubnubyas: You have a point there on the throwing of NumberFormattingException, I will look into adding that too. Thank you :>.

src/main/java/monday/task/TaskList.java
nubnubyas: That's a good suggestion, I will work on it.

src/main/java/monday/task/Task.java
nubnubyas: That's a good suggestion, was feeling that the naming was unnatural too. Thank you!

src/main/java/monday/storage/Storage.java
nubnubyas: Thanks for the suggestion, couldn't really think of a good name back then, thanks for providing a alternative naming.

src/main/java/monday/storage/Storage.java
nubnubyas: Good catch, did not realise the redundant try block. Thank you!

PR #313, opened 2023-08-29
[andrechuakj] iP

andrechuakj:
# Your Chatbot
> "When you fall, I'll be there" - Floor (Source: _trust me_)
Your Chatbot frees your mind so that you have more space for important things like Djikstra algorithm. It,
- doesn't have a GUI so that you can train your typing speed 🥇 
- doesn't need a GUI because a true coder uses CLI 😎 
- uses OOP 💯 

All you need to do is,
1. Download it from 👉 [here](https://youtu.be/dQw4w9WgXcQ?si=pT5heV9Bpg1ZJyEA). 👈 
2. Double click it and pray 🙏 that the code runs.
3. Add your tasks ✅ 
4. Procrastinate 
5. ??
6. Profit 💵 

And it is **FREE**!

Other features
- [X] O(1) adding task
- [X] O(n) deleting task (will be optimized in the future)
- [X] Only has v0.1
- [ ] Bugs (coming soon)
- [ ] Funny description (work in progress)
---
If you are a Java programmer or if you are not, you can use Your Chatbot to practice Java too!! Here's the `main` method:
```java
public class Duke {
    public static void main(String[] args) throws DukeException {
        new Duke("data/duke.txt").run();
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,259 @@
...
+        if (!createFile(f)) {
+            try {
+                loadData(f, list);
+            } catch (DukeException e) {
+                throw e;
+            }
+        }
+
+        boolean end = false;
+        while (!end) {
```
nixonwidjaja: Perhaps consider abstracting your if else or switch case statements so there's not much indentation here?

nixonwidjaja (changes requested):
Great job on making your code simple and concise! Perhaps consider abstracting away some logics in your Duke.java into different classes to make your code more modular and readable. I'm not sure if this is your final code at this moment since I don't see any packages and Gradle is not integrated yet. Overall, good work!

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,18 @@
+package duke;
+
+/**
+ * Custom exception class for Duke application.
+ * DukeException is thrown when there are errors or exceptions specific to the Duke application.
+ */
+public class DukeException extends Exception {
```
ryamgoh, line 7: Hi Andre, code looks hella clean!

I think it will also be nice to segment your code into packages for nicer file structure

Something similar to this:
image

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,170 @@
...
+    private static final String LINE = "──────────────────────────────"
+            + "────────────────────────────────────────────";
+    private static final String LOGO = " _____   __                 _____ _           _   _           _  ___\n"
+            + "|  _\\ \\ / /                /  __ \\ |         | | | |         | ||_  |\n"
+            + "| |  \\ V /___  _   _ _ __  | /  \\/ |__   __ _| |_| |__   ___ | |_ | |\n"
+            + "| |   \\ // _ \\| | | | '__| | |   | '_ \\ / _` | __| '_ \\ / _ \\| __|| |\n"
+            + "| |   | | (_) | |_| | |    | \\__/\\ | | | (_| | |_| |_) | (_) | |_ | |\n"
+            + "| |_  \\_/\\___/ \\__,_|_|     \\____/_| |_|\\__,_|\\__|_.__/ \\___/ \\__|| |\n"
+            + "|___|                                                           |___|\n";
+    private static final String greetPhrase = LINE
```
ryamgoh: A final variable should be in capital case! :)

greetPhrase => GREETPHRASE

Likewise for SENDOFFPHRASE

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param commandDetails The list of command details, which may include task information.
+     * @param taskType       The type of task to add (e.g., "T" for Todo, "D" for Deadline, "E" for Event).
+     */
+    public AddCommand(ArrayList<String> commandDetails, String taskType) {
+        super(commandDetails);
+        this.taskType = taskType;
+    }
+
+    @Override
```
ryamgoh, line 33: I think can still write a javadoc for these execute statements since they have differing behaviours across the different classes! :)

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,29 @@
...
+import duke.DukeException;
+import duke.Storage;
+import duke.Ui;
+import duke.task.Task;
+import duke.task.TaskList;
+
+import java.util.ArrayList;
+
+public class FindCommand extends Command {
+
```
ryamgoh, line 16: javadocs for constructors? In case got overloaded constructor

src/main/java/Duke.java
andrechuakj: Thanks for the feedback, I've taken that into consideration and updated my code with more abstraction.

PR #317, opened 2023-08-29
[LicongHuang] iP

LicongHuang:


src/main/java/cheese/Cheesebot.java
```diff
@@ -0,0 +1,127 @@
...
+    ui.showBye();
+    scanner.close();
+  }
+  
+  /**
+   * Processes the user input.
+   *
+   * @param input The user input.
+   */
+  private void processCommand(String input) {
```
Fallman2, line 86: Would be more efficient to make use of polymorphism by creating command objects that execute themselves with a method

src/main/java/cheese/Parser/Parser.java
```diff
@@ -0,0 +1,229 @@
...
+public class Parser {
+
+    /** Regex patterns for parsing user input */    
+    final static String byPatternString = "\\(by: (.*?)\\)";
+    final static String fromPatternString = "from: (.*?) to:";
+    final static String toPatternString = "to: (.*)\\)$";
+
+  final static Pattern byPattern = Pattern.compile(byPatternString);
+  final static Pattern fromPattern = Pattern.compile(fromPatternString);
+  final static Pattern toPattern = Pattern.compile(toPatternString);
```
Fallman2: All above variables should be in Screaming Snake Case. It might also be better to directly input the strings into the methods to initiate the Patterns instead if the strings are only used once.

src/main/java/cheese/Parser/Parser.java
```diff
@@ -0,0 +1,229 @@
...
+        toMatcher.find();
+        return toMatcher;
+    }
+
+    /**
+     * Converts user input to LocalDate or null
+     * @param dateInput User input
+     * @return LocalDate
+     */
+    public LocalDate dateTimeConverted(String dateInput) {
```
Fallman2: Method name should start with a verb, i.e. convertDateTime

src/main/java/cheese/Parser/Parser.java
```diff
@@ -0,0 +1,229 @@
...
+        DateTimeFormatter inputformat = DateTimeFormatter.ofPattern("dd-MM-yyyy");
+        DateTimeFormatter outputformat = DateTimeFormatter.ofPattern("MMM dd yyyy");
+        Optional<LocalDate> localDate = parseDate(dateInput, inputformat);
+        if (localDate.isPresent()) {
+            return localDate.get();
+        } else {
+            LocalDate localDateMMM = parseMMMFormat(dateInput, outputformat);
+
+
+  public LocalDate dateTimeConverted(String dateInput) {
```
Fallman2: Same as above

src/main/java/cheese/Parser/Parser.java
```diff
@@ -0,0 +1,229 @@
...
+        if (inputSplit.length < 2 && isCommand(input)) {
+          throw new IllegalArgumentException("☹ OOPS!!! The description of a " + command + " cannot be empty.");
+        } else if (inputSplit.length < 2 && !isCommand(input)) {
+          throw new IllegalArgumentException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
+        break;
+
+        case "event":
+        String[] eventInfo = taskDescription.split(" /from ", 2);
+        String[] eventInfo2 = eventInfo[1].split(" /to ", 2);
```
Fallman2: Numbers should not be used in variable names

src/main/java/cheese/Storage/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+   * Loads the tasks from the file and returns a TaskList object
+   * @return TaskList object
+   */
+  public TaskList loadTask() {
+    try {
+      Scanner sc = new Scanner(this.file);
+      while (sc.hasNextLine()) {
+        String line = sc.nextLine();
+        String[] splitLine = line.split("\\|");
+        char type = splitLine[0].charAt(1);
```
Fallman2: A more descriptive name like taskType might be better

src/main/java/cheese/Task/Task.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Represents a task
+ */
+public class Task {
+    private char type;
+    private String description;
+    private String dateTime1;
+    private String dateTime2;
+    private LocalDate dateTimeHr1;
+    private boolean isDone;
```
Fallman2: Here, all 3 tasks share the same type which may lead to less efficient code with lower visual clarity. It may be better to split the three tasks into their own classes.

src/main/java/cheese/Ui/Ui.java
```diff
@@ -0,0 +1,66 @@
...
+      "\n\tdelete <task number>,"+
+      "\n\tmark <task number>,"+
+      "\n\tfind <keyword>,"+
+      "\n\tbye");
+
+    System.out.println("-----------------------------------------------");
+  }
+
+  /**
+   * show error message
```
Fallman2: Probably meant bye message

Fallman2 (commented):
Visual clarity could be improved by splitting some classes into subclasses (such as commands and tasks) but apart from that, looks good overall.

PR #318, opened 2023-08-29
[jacobcuison] iP

jacobcuison:
# Duke Max
> "Trust the commute." - [Mikel Arteta, 2023](https://sports.yahoo.com/mikel-arteta-arsenal-boss-compares-151156611.html?src=rss)

Duke Max helps you keep track of
* tasks
* deadlines
* events

All it takes is:
1. Opening the jar file
2. Add tasks
3. Mark when complete

It's _**that**_ simple! ⭐

It uses Java as well. Here's the `main` method:
```java
    new Max("./data/max.txt").run();
}
```
---
GFMD PR Task:

- [x] a heading
- [x] a bullet list
- [x] a numbered list
- [x] a task list
- [x] an emoji
- [x] a blockquote
- [x] a hyper link
- [x] inline code
- [x] some text formatting: bold, italic, stikethrough etc.


src/main/java/max/Max.java
```diff
@@ -0,0 +1,214 @@
...
+//     */
+//    private void handleUserInput() {
+//        Label userText = new Label(userInput.getText());
+//        Label dukeText = new Label(getResponse(userInput.getText()));
+//        dialogContainer.getChildren().addAll(
+//                DialogBox.getUserDialog(userText, new ImageView(user)),
+//                DialogBox.getDukeDialog(dukeText, new ImageView(max))
+//        );
+//        userInput.clear();
+//    }
```
alyssaongyx: Perhaps it will be better to not keep large chunks of commented code

src/main/java/max/Max.java
```diff
@@ -0,0 +1,214 @@
...
+                String fullCommand = ui.readCommand(); // return the first word of input
+                Command c = parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
+            } catch (MaxException e) {
+                ui.showError(e.getMessage());
+            } catch (DateTimeParseException e) {
+                ui.showError("Please use yyyy-mm-dd format!");
+            } finally {
+                ui.showLine();
```
alyssaongyx: I like your single level of abstraction here

src/main/java/max/commands/AddCommand.java
```diff
@@ -0,0 +1,49 @@
+package max.commands;
+
+import max.commands.*;
+import max.exception.MaxException;
+import max.parser.Parser;
+import max.storage.Storage;
+import max.tasks.TaskList;
+import max.tasks.*;
```
alyssaongyx: Perhaps it would be better to import classes explicitly?

src/main/java/max/commands/UnmarkCommand.java
```diff
@@ -0,0 +1,48 @@
...
+import max.storage.Storage;
+import max.tasks.TaskList;
+import max.tasks.Task;
+import max.ui.Ui;
+
+/**
+ * Represents list command.
+ */
+public class UnmarkCommand extends Command {
+    public static final String COMMAND_WORD = "unmark";
```
alyssaongyx, line 14: Would it be possible to make the variable private?

src/main/java/max/parser/Parser.java
```diff
@@ -0,0 +1,189 @@
...
+            return new ListCommand();
+
+                case MarkCommand.COMMAND_WORD:
+                    return handleMark(fullCommand);
+                case UnmarkCommand.COMMAND_WORD:
+                    return handleUnmark(fullCommand);
+                case DeleteCommand.COMMAND_WORD:
+                    return handleDelete(fullCommand);
+                case FindCommand.COMMAND_WORD:
+                    return handleFind(fullCommand);
```
alyssaongyx: Perhaps it would be better if they have the same indentation

src/test/java/max/tasks/TaskListTest.java
```diff
@@ -0,0 +1,27 @@
+package max.tasks;
+
+import max.exception.MaxException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.junit.jupiter.api.Test;
+public class TaskListTest {
+    @Test
+    public void delete_taskNumberLargerThanMax_exceptionThrown() {
```
alyssaongyx, line 10: Good job on test case naming!

alyssaongyx (commented):
LGTM😃, the codebase is readable with good coding standards applied. There are proper abstractions used and I see that you did optional tasks such as ENUMS as well so good job on that. There are not much issues overall and I enjoyed reading your codebase.

src/main/java/max/commands/CommandEnum.java
```diff
@@ -0,0 +1,5 @@
+package max.commands;
+
+public enum CommandEnum {
```
s0ngyang, line 7: I like the use of enums here.

src/main/java/max/commands/DeleteCommand.java
```diff
@@ -0,0 +1,48 @@
+package max.commands;
+import max.commands.*;
```
s0ngyang: Perhaps you could clean up your imports across all files?

src/main/java/max/parser/Parser.java
```diff
@@ -0,0 +1,189 @@
...
+
+        case ExitCommand.COMMAND_WORD:
+            return new ExitCommand();
+
+                default:
+                    throw new MaxException("Invalid command sir.");
+            }
```
s0ngyang: Could this be the same indentation as the other top level branches?

src/main/java/max/storage/Storage.java
```diff
@@ -0,0 +1,115 @@
...
+            String to = task.substring(toIndex + 3).trim();
+            LocalDate fromDate = LocalDate.parse(from);
+            LocalDate toDate = LocalDate.parse(to);
+            tasks.add(new Event(item, fromDate, toDate, isDone));
+            break;
+        }
+        default:
+            throw new MaxException("Hey! Your storage file is corrupted.");
+        }
+    }
```
s0ngyang: Consider replacing the "magic numbers" such as 1, 3, 5 and 7.  Maybe `String.split()` would be more useful here than `substring()`.

s0ngyang (commented):
LGTM! Overall this is a well-organised codebase which was easy to understand!

src/main/java/max/storage/Storage.java
jacobcuison: Thanks for the feedback! I will update this in future commits :)

src/main/java/max/commands/UnmarkCommand.java
jacobcuison: Thanks for pointing this out! Perhaps I will implement a public `getCommandWord()` method instead. Do you think this would be better?

alyssaongyx: Yup, I think this would be better. Though, I was thinking if you could use enums?

PR #319, opened 2023-08-29
[limjunxian1] iP

limjunxian1:
# HelpBuddy
> “Not only should you have a to-do list, but it must become your best friend.” — Jim Kwik ([Source](https://medium.com/@oaklinejournal/25-quotes-from-successful-people-on-the-power-of-writing-to-do-lists-8a0b3729a34c))

HelpBuddy will be your new best buddy. It saves a list of tasks you need to do. It's,
* text-based
* easy to use
* ~~FAST~~ ***SUPER*** FAST

All you need to do is,
1. download it from [here](https://github.com/limjunxian1/ip/releases/download/v0.1/helpbuddy.jar)
2. open your CLI and navigate to the folder where the `helpbuddy.jar` file is
3. run the command `java -jar helpbuddy.jar`
4. add your tasks 
5. allow it to enhance your task organization :grinning:

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Save deadlines
- [ ] Reminders (coming soon)
#
```ruby
public class Main {
        new HelpBuddy("data/tasks.txt").run();
    }
}
```


src/main/java/helpbuddy/command/Command.java
```diff
@@ -0,0 +1,40 @@
+package helpbuddy.command;
```
Seonlo99, line 1: I think it's really great that you implemented the stretch goal of splitting each of your commands into another class.

src/test/java/helpbuddy/ui/ParserTest.java
```diff
@@ -0,0 +1,115 @@
...
+import helpbuddy.command.Command;
+import helpbuddy.command.DeleteCommand;
+import helpbuddy.command.MarkCommand;
+import helpbuddy.command.UnmarkCommand;
+import helpbuddy.exception.HelpBuddyException;
+import helpbuddy.task.Deadline;
+import helpbuddy.task.Event;
+import helpbuddy.task.ToDo;
+
+public class ParserTest {
```
Seonlo99, line 21: I think it is good that you have cover a variety of tests. it might be good to also include test for invalid input such as accessing an index that is not on the list.

src/main/java/helpbuddy/ui/Ui.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param taskList the taskList that stores Task objects.
+     */
+    public void printListMessage(TaskList taskList) {
+        if (taskList.isEmpty()) {
+            printMessageBlock("You have no task in your list.\n");
+        } else {
+            String messageOutput = "Here are the tasks in your list:\n";
+            for (int i = 0; i < taskList.getSize(); i++) {
+                int taskIndex = i + 1;
+                messageOutput += "\t" + taskIndex + ". " + taskList.getTask(i) + "\n";
```
Seonlo99: I think it might be better if the taskList class could implement Iterable interface so that you can use for each loop instead of using getTask(i).

src/main/java/helpbuddy/ui/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+            String[] taskDateDetails = taskDetails[1].split("/to", 2);
+            String startTime = taskDateDetails[0].trim();
+            if (startTime.isBlank()) {
+                return new AddCommand(new Event(taskName, null, null));
+            }
+            if (taskDateDetails.length == 1) {
+                return new AddCommand(new Event(taskName,
+                        LocalDateTime.parse(startTime, formatter),
+                        null
```
Seonlo99: I think it is a really good idea to allow the creating of classes and let the respective classes handle the error when the user has missing inputs.

Seonlo99 (commented):
LGTM.

PR #347, opened 2023-08-29
[AprupKale] iP

AprupKale:
# Max
## your most efficient manager 🏅 

> “Your mind is for having ideas, not holding them.” – David Allen

Max is
- easy to learn
- super fast
- SUPER *AWESOME*

What you need to do is
- [ ] fork this repo from [here](https://github.com/AprupKale/ip/)
- [ ] clone your fork locally
- [ ] run the bot locally
- [x] HAVE A FUN TIME BEING *ORGANIZED*

What kind of tasks can you add
1. Todo
2. Deadline
3. Event 

**THE BEST PART IS IT IS FREE**

All you need to do to run the bot is call `Duke.main();`

```java
    Duke max = new Duke("data/duke.txt");
    max.run();
}
```


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,102 @@
...
+                for (int i = 0; i < this.getNumberOfTasks(); i++) {
+                    if (!(taskList.getTaskAt(i).equals(this.getTaskAt(i)))) {
+                        return false;
+                    }
+                }
+                return true;
+            } else {
+                return false;
+            }
+        }
```
songfangyl: Is there any other way to do this to avoid this deep nesting?
Here is my suggestion:

```
if (taskList.getNumberOfTasks() == this.getNumberOfTasks()) {
    return false;
}
for (int i = 0...) {
    ....
}
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Gets user input.
+     *
+     * @param taskList TaskList object with list of tasks.
+     * @param storage Storage object to interact with data file.
+     * @throws DukeException If any exception is encountered.
+     */
+    public void getUserInput(TaskList taskList, Storage storage) throws DukeException {
```
songfangyl: I think the name of this method is not so accurate, because when I see "get" in method name, I always expect this method return something to me, but this method does not return anything 🤔

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+                    display("Got it. I've added this task:\n" + add.toString()
+                            + "\nNow you have " + taskList.getNumberOfTasks() + " tasks in the list.");
+                } catch (NullPointerException e) {
+                    throw new DukeException("OOPS!!! Could not add task to the list");
+                }
+            }
+
+            storage.writeToFile(taskList);
+        }
+    }
```
songfangyl: It is better to split this method into multiple methods, because this method not only get user input, but also display the message. My suggestion is the display message can be done in another method, an example shows below

```ruby
public void printAddTaskMessage(Task task, TaskList taskList) {
        System.out.println("Got it, I've added this task:\n    "  +
                task.toString() + "\n" +
                "Now you have " + taskList.size() + " tasks in the list.");
}
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Draws horizontal lines.
+     */
+    public static void lines() {
+        System.out.println("");
+        for (int i = 0; i < 50; i++) {
+            System.out.print("-");
+        }
+        System.out.println("");
+    }
```
songfangyl: I like the way you print the line! I just realized I am hard coding the line🤣

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,99 @@
...
+     * Returns status of task as String.
+     *
+     * @return Status of the Task.
+     */
+    public String getStatusIcon() {
+        return (this.isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Return status of task as Integer.
```
songfangyl: Should be "Returns" instead of "Return" right?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+     * Displays greeting message.
+     */
+    public void greet() {
+        display("Hello! I'm Max\n" + "What can I do for you?");
+    }
+
+    /**
+     * Displays exit message.
+     */
+    public void exit() {
```
songfangyl: Is it better to have a method name like showExitMessage? 🤔

songfangyl (commented):
Overall is good, I think for the UI object, it did more than just interact with the user, which is not so accurate, but I am inspired by your code, after reveiew your code, I realized that I got a lot of things can be improved! 😃

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+                }
+                String[] params = line.split(" \\| ");
+                boolean isDone = params[1].equals("1");
+                switch (params[0]) {
+                case "T":
+                    list.addToList(new Todo(params[2], isDone));
+                    break;
+
+                case "D":
+                    list.addToList(new Deadline(params[2], isDone, LocalDateTime.parse(params[3], Duke.TIME_FORMAT)));
```
yezkez10: Maybe you want to place line breaks to improve readability?
```suggestion
                    list.addToList(new Deadline(params[2], isDone,
                        LocalDateTime.parse(params[3], Duke.TIME_FORMAT)));
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+                    taskList.addToList(add);
+                    display("Got it. I've added this task:\n" + add.toString()
+                            + "\nNow you have " + taskList.getNumberOfTasks() + " tasks in the list.");
+                } catch (NullPointerException e) {
+                    throw new DukeException("OOPS!!! Could not add task to the list");
+                }
+            }
+
+            storage.writeToFile(taskList);
+        }
```
yezkez10: I like the very clear logical flow here, but is there a way you can shorten the code here by abstracting each 'case' using the enum class?

src/main/java/duke/exceptions/DukeDateTimeParseException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exceptions;
+
+public class DukeDateTimeParseException extends DukeException {
+    public DukeDateTimeParseException() {
+        super("OOPS!!! Date and Time format does not match dd mmm yyyy - hh:mm");
+    }
```
yezkez10, line 9: Great abstraction of exception! Maybe you want to articulate the problem clearer by giving an example of a date and time that matches the format dd mm yyyy - hh:mm. Maybe the message can be:
"OOPS!!! Date and Time format invalid! Try with input format 30 Jul 2003 2:00"

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,102 @@
...
+        for (int i = 0; i < this.list.size(); i++) {
+            s.append((i + 1) + "." + this.list.get(i));
+            if (i < this.list.size() - 1) {
+                s.append("\n");
+            }
+        }
+        return s.toString();
+    }
+
+    @Override
```
yezkez10, line 106: Love that you have narrowed down and specified the equality for better precision!
Would be good to add JavaDoc to justify why the overriding equals methods are necessary, ie what make it different from the overarching equals method such that it warrants its own overriding method.

yezkez10 (commented):
LGTM👍 (looks GREAT to me)!
Overall, I like that the code was very OOP-ised. Excellent abstraction of methods into their relevant classes. On top of that, _phenomenal_ abstraction of exceptions and exception handling. 
If the circumstances allow it, you could try truncating the logic flow in getUserInput using the enum class, to make your code _even_ easier to read.
Great work thus far! 
"Be humble. Be hungry. And always be the hardest worker in the room." - Dwayne Johnson

PR #350, opened 2023-08-29
[zekone] iP

zekone:
# Thorndike Premium

> "The chadest of chadbots is Thorndike, try it and win it all." – Sun Tze, Art of War ([source](https://github.com/zekone/ip))

## Features
 - ~~No nonsense~~
 - **Best** cat
 - *AI powered*

## How to get it
1. Visit our [website](https://github.com/zekone/ip).
2. Pay a cheap subcription fee of SGD1,000 per month.
3. Enjoy your very own chad cat.

## How to use it
 Go to the directory that contains Main.java with the `main` method and run:
 ```java
 public class Main {
        System.out.println("GIB MONEY PLZ.");
    }
}
 ```

## Review
Please help us review our product.
 - [ ] Good
 - [ ] Amazing
 - [x] Awesome :smile:


src/main/java/MainWindow.java
```diff
@@ -0,0 +1,57 @@
...
+import exceptions.ThorndikeException;
+import javafx.fxml.FXML;
+import javafx.scene.control.Button;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+
+public class MainWindow extends AnchorPane {
```
jianrong7: perhaps add some javadocs to the class?

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void setDuke(Thorndike d) {
+        thorndike = d;
+    }
+
+    public void greet() {
+        dialogContainer.getChildren().add(
+                DialogBox.getDukeDialog(Message.GREET, dukeImage));
+    }
```
jianrong7: perhaps add javadocs for these methods as well

src/main/java/Thorndike.java
```diff
@@ -0,0 +1,42 @@
...
+    public String getResponse(String input) throws ThorndikeException {
+        Command command = CommandParser.parse(input);
+
+        if (command == null) {
+            throw new InvalidCommandException();
+        }
+
+        return command.execute(taskList);
+    }
+}
```
jianrong7: javadocs here too!

src/main/java/command/CommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+
+            return new CmdAddTask(new Event(description, from, to));
+        }
+
+        if (command.equals("find")) {
+            return new CmdFind(description);
+        }
+
+        return null;
+    }
```
jianrong7: perhaps consider `switch-case` statements instead of `if` statements. but still works good!

jianrong7 (commented):
Good job! I like your code and it is generally very clean and readable. 😸

src/main/java/command/CommandParser.java
zekone: Thanks for the suggestion, I am using `if-return` statements is because the formatting of my IDE for `switch-case` does not comply with the standard of this module. I can change the formatting setting but it will take some time.

src/main/java/Thorndike.java
zekone: Will be adding soon.

src/main/java/MainWindow.java
zekone: will do

src/main/java/MainWindow.java
zekone: will do

src/main/java/command/CommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+        for (String arg : argList) {
+            String option = StringUtility.getFirstWord(arg);
+            String value = StringUtility.removeFirstWord(arg);
+
+            if (!option.equals(command)) {
+                arguments.put("command", command);
+                arguments.put("description", value);
+                arguments.put(option, value);
+            } else {
+
```
Sasmik23: Is the else statement needed?

src/main/java/command/CommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+
+        if (command.equals("list")) {
```
Sasmik23: Could consider the use of enums? E.g.

private static final String COMMAND_LIST = "list";

if (command.equals(COMMAND_LIST)) {
    return new CmdList();

src/main/java/command/CommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+            if (from == null || to == null) {
+                throw new InvalidDateTimeFormat();
+            }
+
+            return new CmdAddTask(new Event(description, from, to));
+        }
+
+        if (command.equals("find")) {
+            return new CmdFind(description);
+        }
```
Sasmik23: Any reason for choosing if statements over else-if? Not sure if it's considered a violation but might be more consistent!

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public abstract class Storage {
+    /** Path to the textfile. */
+    public static final String TASK_FILE_PATH = "data/tasks.txt";
+
+    /** Character indicator that separates one elements in the textfile. */
+    public static final String TASK_FILE_SEPARATOR = "-";
+
+    /** Standard format for time and date. */
+    public static final String TASK_FILE_TB_FORMAT = "dd/MM/yyyy HHmm";
```
Sasmik23, line 27: does it need to be public or can it be changed to private?

Sasmik23 (commented):
Overall great readability and good use of oop!

src/main/java/command/CommandParser.java
zekone: thanks, its a dead code that i forgot to remove.

src/main/java/command/CommandParser.java
zekone: preference and also easier to read. `Else-if` is hard to see the separation between each clauses as each block cannot be separated by a line break.

src/main/java/command/CommandParser.java
zekone: I have considered it but decided not to because the amount of times I used it is quite few, usually 1 or 2 times for each variable. Thanks for the suggestion anyway!

src/main/java/storage/Storage.java
zekone: For now, only `Thorndike` is accessing `TASK_FILE_PATH`.  But since its `static` `final`, other classes can safely access it without risk and I do see the possibility that other classes might need the other two in the future.

PR #362, opened 2023-08-29
[Althea Chua] iP

chuababyy:
# DukePro

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

 - [x] Managing tasks
 - [ ] Reminders (coming soon)

```java
        new Duke(filePath).run();
    }
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Initialises the application, displays a welcome message and
+     * enters a loop to process user inputs until the use enters "bye".
+     */
+    public void run() {
+        Scanner scanner = new Scanner(System.in);
+        ui.showWelcome();
+        boolean isNotBye = true;
+        while (isNotBye) {
+            isNotBye = parser.createTaskAction(scanner.nextLine());
```
butteredyakiimo: I like how you check for the terminating condition here. The use of `isNotBye` makes it quite intuitive to understand when the program should exit.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Returns {@code true} if the start date is before or same as the end date, and
+     * returns {@code false} otherwise.
+     *
+     * @param start Start date to be compared.
+     * @param end End date to be compared.
+     */
+    public boolean isStartDateBefore(LocalDateTime start, LocalDateTime end) {
```
butteredyakiimo: I think a more intuitive name for this could be `isStartBeforeEnd`! What do you think?

src/main/java/duke/FileStorage.java
```diff
@@ -0,0 +1,151 @@
...
+     * Adds all items in the TaskList to the file in the specified file path.
+     *
+     * @param taskList The list of tasks to be saved.
+     */
+    public void saveTasks(TaskList taskList) {
+        File file = new File(filePath);
+        file.mkdirs();
+
+        if (file.exists()) {
+            boolean isDeleteFileSuccess = file.delete();
```
butteredyakiimo: I think instead of deleting and creating a new file each time the tasklist updates, an alternative would be to to use `writer.write(task.toFileString())` instead! This method will overwrite the existing data, allowing you to use the same file for every update :D What do you think?

src/main/java/duke/FileStorage.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    /**
+     * Returns a LocalDateTime object by converting the String provided.
+     *
+     * @param date The date string to be converted.
+     * @return A LocalDateTime object representing the date and time.
+     */
+    public LocalDateTime setDate(String date) {
```
butteredyakiimo: I think a more intuitive name for this method could be `convertToDate`! what do you think?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    /**
+     * Converts a date string to a LocalDateTime object.
+     *
+     * @param date The date string in the specified format.
+     * @return A LocalDateTime object representing the parsed date and time.
+     */
+    public LocalDateTime setDate(String date) {
```
butteredyakiimo: similar to previous comment on the same method in storage class!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    /**
+     * Processes user input to create and execute task-related actions.
+     *
+     * @param input The user input to be processed.
+     * @return {@code true} if the application should continue running. {@code false} if it should exit.
+     */
+    public boolean createTaskAction(String input) {
```
butteredyakiimo: I like the rationale behind this method! I think a more intuitive name for this method could be `isTaskCreated`, so that when false is returned, it means that the task was not created, and the application should stop running. What do you think?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,225 @@
...
+        String[] userInputParts = input.split(" ", 2);
+        String commandUser = userInputParts[0];
+        CommandType command = parseCommand(commandUser);
+
+        switch (command) {
+        case EMPTY:
+            ui.showEmptyMessage();
+            return true;
+
+        case UNKNOWN:
```
butteredyakiimo: I think leaving the `UNKNOWN` case to the last case would make reading this block of code smoother!

butteredyakiimo (commented):
LGTM :+1: Clever use of switch statements and booleans in implementing code logic. Just some minor naming issues to fix!

README.md
xenosf, line 1: It seems like when you moved your classes into packages, the class names ("Duke" here) got auto-reformatted to contain the package name. Since the "Duke" here is not a class name, should these be changed back? (`duke.Duke` -> `Duke`)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,84 @@
...
+    //    }
+
+    //    /**
+    //     * Entry point of the Duke application.
+    //     *
+    //     * @param args Command-line arguments. Not used in this context.
+    //     */
+    //    public static void main(String[] args) {
+    //        new Duke().run();
+    //    }
```
xenosf: Should this unused code be deleted instead of commented out to keep the code cleaner?

src/main/java/duke/components/Parser.java
```diff
@@ -0,0 +1,227 @@
...
+            return CommandType.DELETE;
+        } else if (input.equals("list")) {
+            return CommandType.LIST;
+        } else if (input.equals("")) {
+            return CommandType.EMPTY;
+        } else if (input.equals("find")) {
+            return CommandType.FIND;
+        } else {
+            return CommandType.UNKNOWN;
+        }
```
xenosf, line 65: This if statement is very long -- would it be more readable if it was a switch-case instead?
You can consider using something like [`CommandType.valueOf()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Enum.html#valueOf(java.lang.Class,java.lang.String)) to make the code more concise.

src/main/java/duke/components/Parser.java
```diff
@@ -0,0 +1,227 @@
...
+        case FIND:
+            String[] inputParts = input.split(" ", 2);
+            String keyword = inputParts[1].trim();
+            TaskList list = fullList.findTask(keyword);
+            if (list.getSize() == 0) {
+                return ui.showNoFind();
+            } else {
+                return ui.showFind(list);
+            }
+
```
xenosf, line 184: I like the use of line breaks to separate logical blocks and make the code easier to read!

src/main/java/duke/components/TaskList.java
xenosf, line 1: Should the unused code blocks in this file be deleted instead of commented out to keep the code cleaner? If the old code is needed for reference, it is possible to view past versions using Git

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param description Name of the deadline task.
+     * @param deadline Date before which task should be completed.
+     * @param isDone {@code true} if the task is marked as done, {@code false} otherwise.
+     */
+    public Deadline(String description, LocalDateTime deadline, boolean isDone) {
+        super(description, isDone);
+        this.deadline = deadline;
+    }
+
+    /**
```
xenosf, line 39: This

```java
/**
 * {@inheritDoc}
 */
```

might not be necessary - If the entire comment is inherited from the overridden method, the whole JavaDoc is inherited as well.

PR #365, opened 2023-08-29
[MagnificentCreature] iP

MagnificentCreature:
# Forine 

> “Beep beep.” – Sheep ([source](https://example.com))

Forine **frees** you.

* _test-based_
* _learning is hard_
* _FAST_ ~~SUPER FAST~~ to use

_All you need to do is_,

1. download it from [here](https://github.com/MagnificentCreature/ip).
2. `double-click` it.
3. add your tasks.
4. let it manage your tasks for you :wink:

And it is **FREE**!

## Features:

- [x] Managing words
- [x] Managing boxes with words on its right
- [ ] other things

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/duke/Parser.java [hunk 1]
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param input the user input
+     * @return the Command object
+     * @throws UnknownCommandException if the command is not recognised
+     * @throws MissingDescriptionException if the description is missing
+     * @throws IncorrectCommandFormatException if the command is in the wrong format
+     * @throws InvalidIndexException if the index provided is invalid
+     * @throws InvalidTimeFormatException if the time provided is invalid
+     */
+    public Command dispatch(String input)
```
yihfei, line 98: I like the command class and how it's organised! Perhaps could reconsider renaming this method to parse as I couldn't tell what dispatch meant until I read the javadoc comment

src/main/java/duke/parser/IndexCommandParser.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (NumberFormatException e) {
+            throw new InvalidIndexException(commandName + " index must be a number");
+        }
+        if (index < 0) {
+            throw new InvalidIndexException(commandName + " index cannot be less than 0");
+        }
+    }
+
+    @Override
+    protected Command createCommand(Matcher matcher)
```
yihfei, line 45: I like the use of cases here!

src/main/java/duke/commands/Command.java
```diff
@@ -0,0 +1,21 @@
...
+public interface Command {
+    /**
+     * Executes the command.
+     *
+     * @param taskListStorage The TaskListStorage object that contains the task list.
+     * @throws InvalidIndexException If the index provided is invalid.
+     * @throws MissingDescriptionException If the description is missing.
+     * @throws IncorrectCommandFormatException If the command is in the wrong format.
+     */
+    void execute(TaskListStorage taskListStorage)
```
yihfei: Is there any reason why there wasn't a access modifier here

src/main/java/duke/parser/EventCommandParser.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    //CHECKSTYLE.OFF: LineLength
+    public EventCommandParser() {
+        super("event",
+                "^(?<command>event)(?: (?<description>.*?)?(?<from> /from.*?)?(?<fromTime>\\d{4}-\\d{1,2}-\\d{1,2})?(?<to> /to.*?)?(?<toTime>\\d{4}-\\d{1,2}-\\d{1,2})?)?$");
+    }
+    //CHECKSTYLE.ON: LineLength
+
+    @Override
+    protected void validate(Matcher matcher)
```
yihfei, line 27: I like how comprehensive this method is :))

yihfei (commented):
Overall, the code looks extremely clean and readable. Really nothing much to change here!

src/main/java/duke/TaskListStorage.java
```diff
@@ -0,0 +1,258 @@
...
+            throws FileNotFoundException, IncorrectCommandFormatException, MissingDescriptionException {
+        Scanner sc = new Scanner(this.file);
+        while (sc.hasNextLine()) {
+            String line = sc.nextLine();
+            String[] taskInfo = line.split(" \\| ");
+            String taskType = taskInfo[0];
+            boolean isDone = taskInfo[1].equals("1");
+            String taskDescription = taskInfo[2].replace("\\|", "|"); //Unescape
+            try {
+                switch (taskType) {
```
dylkaw, line 105: Is there a reason why you did not use an enum for the task types?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        /**
+         * Returns the command with the specified string representation.
+         * Idea was taken from https://www.baeldung.com/java-enum-values#locating-values
+         *
+         * @param commandString the string representation of the command
+         * @return the command with the specified string representation
+         */
+        public static Commands valueOfLabel(String commandString) {
```
dylkaw: Maybe this can be named using a verb?

dylkaw (commented):
Overall well organised and easy to read!

src/main/java/duke/commands/Command.java
MagnificentCreature: Oops, fixed

src/main/java/duke/Parser.java
MagnificentCreature: I am considering it, but I already have another method called parse. That parse will be overridden for each command depending on how the command is meant to be parsed. Dispatch is used to run that parse.

That  being the case I could not find a comprehensive way to fuse the two and left it as such

src/main/java/duke/TaskListStorage.java
MagnificentCreature: This is the rawest form of parsing the inputs from the file, as such enums are not necessary. Creating an enum will mean I must then associate the enum with the letter I'm using to save to the file which seems like extra space and computational work for not much increase in understanding/readability of the code. 

Without enums, it should be easy to see D corresponds to creating a deadlineTask since the following line is adding a deadline task to the tasklist.

The only time I might consider using enums is if the task types are a considerably high number that 

1)I cant just use simple alphabets to represent
2)I might need to map what I write in the task file to its own unique task, ie D1, D2 etc.

src/main/java/duke/Parser.java
MagnificentCreature: Thanks for the suggestion! I will change this

PR #370, opened 2023-08-29
[hcs1203] iP

hcs1203:
**Come and Chat over some CarrotCake**

> “Your mind is for having ideas, not holding them.” – [David Allen](https://dansilvestre.com/productivity-quotes)

This app frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
2. double-click it.
3. add your tasks.
5. And it is FREE!

Features:
- [x]  Managing tasks

```
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String div = "____________________________________________________________\n";
+
+        class Task {
```
CJ-Lee01: Should classes be declared inside methods?

CJ-Lee01: I think it is better for you to separate your classes into different files, especially when such class declarations are taking up a significant amount of space in your file.

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+            }
+
+            public String printTask() {
+                return (isDone) ? "[X] " + name + "\n" : "[ ] " + name + "\n";
+            }
+
+            public void addToFile() {
+                try {
+                    FileWriter Writer
+                            = new FileWriter("data/data.txt", true);
```
CJ-Lee01: would it be better to "abstract" away the file path by using a constant declaration instead? It would be painful to refactor if I had to change the file path.

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+        }
+
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>();
+        int count = 0;
+        File file = new File("data/data.txt");
+
+        try {
+            Scanner fileReader = new Scanner(file);
+            while (fileReader.hasNextLine()) {
```
CJ-Lee01: I can see that you are loading the file into the array. However, would it be better to abstract the logic somewhere else?

CJ-Lee01: What if one day maybe the input system change and you are reading from a  database instead?

Or maybe the input is of a different format (e.g. JSON, XML, YAML)

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+                    continue;
+                }
+                for (int i = 0; i < count; i++) {
+                    System.out.print((i + 1) + ". " + tasks.get(i).printTask());
+                }
+                System.out.println(div);
+                input = scanner.nextLine();
+                continue;
+            }
+            if (input.toLowerCase().startsWith("mark ")) {
```
CJ-Lee01: I really like the use of `startWith`.

src/main/java/Duke.java
AprupKale: Does a nested class make logical sense within Duke?

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+                            "  | 0 | " + name);
+                    Writer.close();
+                } catch (IOException e) {
+                    System.out.println("An error has occurred.");
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        class ToDos extends Task {
```
AprupKale: Is ToDo a better name than ToDos as a class is only a template?

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+                    Writer.close();
+                } catch (IOException e) {
+                    System.out.println("An error has occurred.");
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        class Deadline extends Task {
+            LocalDate dueDate;
```
AprupKale: Missing access specifier!

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+                    Writer.close();
+                } catch (IOException e) {
+                    System.out.println("An error has occurred.");
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        class Events extends Task {
+            LocalDateTime startDateTime;
```
AprupKale: Missing access specifiers!

src/main/java/Duke.java
```diff
@@ -1,10 +1,287 @@
...
+        int count = 0;
+        File file = new File("data/data.txt");
+
+        try {
+            Scanner fileReader = new Scanner(file);
+            while (fileReader.hasNextLine()) {
+                String input = fileReader.nextLine();
+                String[] parts = input.split("\\|");
+
+                if (parts[0].trim().equalsIgnoreCase("t")) {
```
AprupKale: Would be better to use an enum and use switch cases instead of if-else!

AprupKale (commented):
LGTM! However you can make separate classes instead of nested classes within Duke.

PR #385, opened 2023-08-30
[scarletblanks] iP

ScarletBlanks:
Till

Till frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/ScarletBlanks/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [X] Managing tasks

```Java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,143 @@
...
+            storage.saveTasksToFile(tasks);
+            System.exit(0);
+            return;
+
+        case "list":
+            Ui.tasksInList(tasks);
+            break;
+
+        case "mark":
+            int number = myScanner.nextInt();
```
xxiaoweii: The naming can be more consistent to improve clarity (number is used here but numero is used at the later part)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,143 @@
...
+            Ui.printMatchingTasks(tasks, inValue);
+            break;
+
+        default:
+
+            inValue += myScanner.nextLine();
+            Ui.unrecognisedCommand();
+            break;
+
+        }
```
xxiaoweii: you might want to close the scanner at the end

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                        input.add(new Deadline(splitted[2], LocalDate.parse(splitted[3]), (splitted[1] == "1")));
+                    } else {
+                        Ui.badDateLoaded();
+                    }
+                    break;
+                }
+                nextLine = bufferedReader.readLine();
+            }
+            bufferedReader.close();
+        } catch (IOException e) {
```
xxiaoweii, line 94: i think it's better to throw different types of error instead of printing the error messages

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,174 @@
...
+        System.out.println(item.toString());
+        this.printLines();
+    }
+
+    /**
+     * Notifies the user that a task has been unmarked
+     *
+     * @param item the item that has been unmarked
+     */
+    public void taskUndone(Task item) {
```
xxiaoweii: The naming of this method can be more specific to suggest that it is a method to print the messages. Same for taskAdd and taskDelete

xxiaoweii (commented):
Overall good effort! Just some minor changes

PR #387, opened 2023-08-30
[Yan Weidong (Evan)] iP

evanyan13:
# Jarvis CLI Task Manager
> "Track and Manage your tasks at your CLI"

Screenshot 2023-09-10 at 5 55 51 PM

**Hi Master! Jarvis is here at your service. Jarvis is a CLI-based Task Manager for:**
- Text-based
- Easy to Learn
- Fast to Use

### All you need to do is: 
1. Download it from [this link](https://github.com/evanyan13/ip/releases/tag/A-Jar)
2. Run java -jar jarvis.jar in the directory where the file is present
3. Add your tasks.
4. Let it manage your tasks for you 😁

### Features:
- [x]  Managing Tasks
- [x]  Filter Tasks based on keyword
- [ ]  Managing Deadlines (coming soon)

Here's the `main` method which can be founded in `src/main/java/jarvis/Jarvis.java`

```java
        Jarvis jarvis = new Jarvis();
        jarvis.start();

        String userInput;

        try {
            while (true) {
                userInput = scanner.nextLine();
                jarvis.respond(userInput);
            }
        } finally {
            scanner.close();
        }
    }

```


src/main/java/jarvis/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Parses a string representation of a task into a Task object.
+     *
+     * @param line The string representation of the task.
+     * @return A Task object representing the parsed task.
+     * @throws InvalidTaskFormatException If the task format is invalid.
+     */
+    public static Task parseStringToTask(final String line) throws InvalidTaskFormatException {
```
tanyyyming: If I read correctly, this should be the method that reads from the data file and converts the string to task? If so, would it be better to name it `decodeStringToTask` to avoid confusion with the method that parses the user input?
```suggestion
    public static Task encodeStringToTask(final String line) throws InvalidTaskFormatException {
```

src/main/java/jarvis/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            String[] lineSplit = line.split("\\|");
```
tanyyyming: Is it better if the variable name `lineSplit` is replaced with `splittedLine`?

src/main/java/jarvis/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+                return new Deadline(taskDetails, formattedDeadlineBy, isCompleted);
+            case "E":
+                String fromTime = lineSplit[3].trim();
+                String toTime = lineSplit[4].trim();
+                LocalDateTime formattedFromTime = parseDateTime(fromTime);
+                LocalDateTime formattedToTime = parseDateTime(toTime);
+                return new Event(taskDetails, formattedFromTime, formattedToTime, isCompleted);
+            default:
+                throw new InvalidTaskFormatException("Unknown Task Type: " + taskType);
+            }
```
tanyyyming: Good naming of local variables in this section!

src/main/java/jarvis/commands/DeleteCommand.java
```diff
@@ -0,0 +1,45 @@
...
+     * list.
+     *
+     * @param taskList The TaskList containing the tasks.
+     * @param ui       The Ui for user interface interactions.
+     * @param storage  The Storage for saving tasks.
+     * @throws InvalidIndexException If an invalid index is provided.
+     */
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws InvalidIndexException {
+        String[] userInputSpilt = userInput.split(" ");
```
tanyyyming: Same here. I don't know whether `userInputSplit` is a good name for a variable since it sounds like a verb (method).

src/main/java/jarvis/tasks/Deadline.java
```diff
@@ -0,0 +1,31 @@
...
+import java.time.format.DateTimeFormatter;
+
+import jarvis.Ui;
+
+/**
+ * Represents the "Deadline" task in Jarvis app.
+ */
+public class Deadline extends Task {
+
+    private LocalDateTime dueDate;
```
tanyyyming: Maybe rename to `dueDateTime` to be more consistent with the other variables such as `toDateTime` and `fromDateTime`?
```suggestion
    private LocalDateTime dueDateTime;
```

src/main/java/jarvis/tasks/Task.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param keyword The keyword the user wish to search for.
+     * @return True if the task title contains the keyword, false otherwise.
+     */
+    public boolean contains(String keyword) {
+        if (title.contains(keyword)) {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
tanyyyming: Since this is a boolean method, is it better if you prefix it with keywords like `has`? Maybe something like `hasTasksWithKeyword`? Though I know that the string method `contains` also returns a boolean value, so its definitely up to you!
And by the way, the method can be simplified.
```suggestion
    public boolean hasTaskWithKeyword(String keyword) {
        return title.contains(keyword);
    }
```

tanyyyming (commented):
In general, the naming standards are followed pretty closely. Good job!

I've pointed out some minor modifications which are not serious offences of the standard. It is up to you to change it or keep it!

src/main/java/jarvis/Parser.java
evanyan13: Good point! Thank you and I have implemented the changes

PR #395, opened 2023-08-30
[xyT-T] iP

xyT-T:
# Slay!

> "Work. Play. Slay."

Slay your tasks and slay your day with the help of _Slay!_ Slay helps you to:

- [x] manage tasks
- [x] remember deadlines
- [ ] organize your work

It's,

- text-based
- easy to use
- fast
- ~~cute~~ SUPER CUTE to interact with :D

All you need to do is,

1. Download it from [here](https://github.com/xyT-T/ip/releases).
2. Do a double-click.
3. Add your tasks.
4. Have fun with it and **SLAY YOUR DAY** 🦖

```ruby
public class Slay {
        new Slay().run();
    }
}
```


src/main/java/Tong.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Scanner;
+import java.util.Arrays;
+
+public class Tong {
+    public static void main(String[] args) {
```
trgao: Perhaps you should abstract out some methods so that the main method is not so long and difficult to read?

src/main/java/Tong.java
```diff
@@ -0,0 +1,132 @@
...
+
+            currentLine = input.nextLine();
+
+            while (!currentLine.equalsIgnoreCase("bye")) {
+                String[] split = currentLine.split(" ");
+                String command = split[0];
+                int order;
+
+                switch (command) {
+                    case "list":
```
trgao: Perhaps you can remove the indentation for the case statements? (switch and case are at the same level of indentation)

trgao (commented):
Perhaps you can abstract out the code into different classes, but otherwise LGTM.

src/main/java/tong/Parser.java
```diff
@@ -0,0 +1,164 @@
+package tong;
+
+import tong.command.*;
```
Song-Mengfei: It might be better to import every classes seperately

src/main/java/tong/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+            return new IncorrectCommand("Sorry, I don't know this type of tasks:(\n"
+                    + "You may choose from what I have:D\n"
+                    + AddCommand.MESSAGE_USAGE);
+        }
+
+        String type = matcher.group("type");
+        String description = matcher.group("description");
+
+        switch (type) {
+            case ToDo.TYPE:
```
Song-Mengfei: There is no indentation for case clauses according to our coding standard 😥
I didn't realise it as well

src/main/java/tong/Ui.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public void showGoodbyeMessage() {
+        out.println("Bye.");
+    }
+
+    /**
+     * Shows the result of a command execution to the user. Includes additional formatting to demarcate different
+     * command execution segments.
+     */
```
Song-Mengfei: I think the method parameters and return value should be involved in JavaDoc.
Such as:

/**
* This method is for ...
*
* @param name ...
* @return returnValue ...
*/

src/main/java/tong/command/UnmarkCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public CommandResult execute() {
+        try {
+            Task toUnmark = taskList.getTask(targetVisibleIndex);
+            taskList.unmarkTask(toUnmark);
+            return new CommandResult(String.format(MESSAGE_UNMARK_TASK_SUCCESS, toUnmark));
+        } catch (TaskNotFoundException e) {
+            return new CommandResult("You don't seem to have so many tasks. " +
```
Song-Mengfei: It might be more clear if you break before the operator "+"

src/main/java/tong/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+            Pattern.compile("(?<description>.+)\\s+"
+                    + "/by\\s+"
+                    + "(?<deadline>\\S+\\s+\\S+)$");
+
+    public static final Pattern EVENT_DES_FORMAT =
+            Pattern.compile("^(?<description>.+)\\s+"
+                    + "/from\\s+"
+                    + "(?<start>\\S+\\s+\\S+)\\s+"
+                    + "/to\\s+"
+                    + "(?<end>\\S+\\s+\\S+)$");
```
Song-Mengfei: I like the way you used Pattern and final value 😀

Song-Mengfei (commented):
Good job! I have learnt a lot!

PR #418, opened 2023-08-30
[rayyan35p] iP

rayyan35p:
# ChatBob 👷‍♂️ 
> _Can we fix it? Yes, we can!_
- Did some work but made mistakes?
- Think you can't fix your mistakes?
- Want to be told that you **CAN** fix it?
- If you answered yes to any of the above, you're in the right place!

### Bob can:

1. Keep track of your tasks
2. And if you ask `canWeFixIt()`, he will return true.

### Look at all this code he wrote!

```java
public String getFirstInput() {
        System.out.println("Hello! I'm Bob");
        System.out.println("What can I do for you?");
        return this.getInput();
    }
```

- [x] Keep track of your tasks
- [ ] Tell you that you can fix it

So in the meantime, watch [this ](https://www.youtube.com/watch?v=-P2-Nmk1uFQ)while waiting for construction to complete!


src/main/java/Duke.java
```diff
@@ -1,10 +1,38 @@
...
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Ui ui = new Ui();
+        Storage storage = new Storage();
+        TaskList list = new TaskList(storage.getFile());
+        list.open();
+        Parser parser = new Parser();
```
dishenggg: Maybe you can consider initializing all your variables in a section on top and leaving a line before running any logic?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,114 @@
...
+            if (inputSplit.length == 1) {
+                throw new DukeException("Input the task number");
+            } else if (inputSplit.length > 2) {
+                throw new DukeException("Invalid command");
+            }  else {
+                return Command.MARK;
+            }
+        case UNMARK:
+            if (inputSplit.length == 1) {
+                throw new DukeException("Input the task number");
```
dishenggg: Perhaps you can consider using a named constant for `"Input the task number"` and `"Invalid command"`?

src/main/java/data/task/TaskList.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Calls a specific method with the specified input based on the specified command.
+     *
+     * @param command The command to be run.
+     * @param input The input string for the command.
+     * @throws DukeException If an EventTask is instantiated with invalid dates.
+     */
+    public void executeCommand(Command command, String input) throws DukeException {
```
dishenggg: I like that you delegated the logic of each command to their own function. It reduced the overall length of the methods

src/main/java/Duke.java
```diff
@@ -1,10 +1,38 @@
...
+
+        String input = ui.getFirstInput();
+
+        while (true) {
```
dishenggg: Well structured code, no deep nesting 👍

src/main/java/data/task/TaskList.java
```diff
@@ -0,0 +1,296 @@
...
+        case DEADLINE:
+            input = input.substring(9);
+            inputArr = input.split(" /by ");
+            taskToAdd = new DeadlineTask(inputArr[0], inputArr[1]);
+            break;
+        case EVENT:
+            input = input.substring(6);
+            inputArr = input.split(" /from ");
+            String description = inputArr[0];
+            inputArr = inputArr[1].split(" /to ");
```
dishenggg: Maybe you can consider using different variables names instead of overwriting them?

dishenggg (commented):
LGTM! Good code quality throughout

hjoneweek (commented):
It seems like there weren't any coding standard violations (that I could find) in the program. I found your code easy to read as well. Looks good to me!

src/main/java/Duke.java
rayyan35p: Yeah that would be neater, thanks!

src/main/java/parser/Parser.java
rayyan35p: Yep ok makes sense

src/main/java/data/task/TaskList.java
rayyan35p: Will do, thanks!

PR #432, opened 2023-08-30
Heran IP

Heran9:
### **Moss ChatBot**

> This is where everything starts😀

Moss can help you remember things. It's

- text-based
- simple
- easy to use

It contains commands such as:

1. todo, deadline, event (to add tasks to your task list) ``
2. list (check your task list) 
example of the implementaion of list :
`if (message.equals("list")) {
            // List all tasks
            System.out.println("____________________________________________________________");
            System.out.println("Here are the tasks in your list:");
            for (int i = 0; i < things.size(); i++) {
                System.out.println(i + 1 + "." + things.get(i).toString("x"));
            }
            System.out.println("____________________________________________________________");`
3. mark (mark things) you have already done
...

Here is the basic implementation of Moss:
```
public class Moss {
    /**
     * The list of tasks managed by the application.
     */
    static ArrayList<Task> things = new ArrayList<>();

    /**
     * The main method that initializes the application and handles user input.
     *
     * @param args Command-line arguments (not used in this application).
     * @throws MossException If there's an issue with loading tasks.
     */
    public static void main(String[] args) throws MossException {
        Storage storage = new Storage();

        try {
            things = (ArrayList<Task>) storage.loadTasks();
        } catch (Exception e) {
            throw new MossException("Failed to load tasks");
        }

        UI ui = new UI();
        ui.greeting();

        Scanner sc = new Scanner(System.in);
        String message = sc.nextLine();

        // Process user input until "bye" is entered
        while (!message.equals("bye")) {
            Parser parser = new Parser(message, things, storage);
            parser.execute(message, things, storage);
            message = sc.nextLine();
        }
        // Farewell message
        ui.bye();
    }

}
```
It is still a base model, and needs further development:

- [ X ] Base implementation
- [ ] Further improvement (still thinking about it)

Can reference [here](https://github.com/Heran9/ip/tree/master/src/main/java/Moss) to see the code implementation.


src/main/java/Moss/Task.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Marks the task as done and prints a confirmation message.
+     */
+    public void markDone() {
+        this.isDone = true;
+        System.out.println("____________________________________________________________");
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println(this.toString());
+        System.out.println("____________________________________________________________");
```
jx124: Perhaps the printing can be handled by the UI class?

src/main/java/Moss/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Processes and executes user commands related to task management.
+     *
+     * @param message The user command to be processed.
+     * @param things The list of tasks being managed.
+     * @param storage The storage object for saving tasks.
+     * @throws MossException If there's an issue with processing commands or managing tasks.
+     */
+    public static void command(String message, ArrayList<Task> things, Storage storage) throws MossException {
```
jx124: Could use a verb for the name of this function since `command` by itself might be ambiguous. Perhaps `executeCommand`?

src/main/java/Moss/Task.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @return The formatted string.
+     */
+    public String toString() {
+        return getStatusIcon() + " | " + this.description;
+    }
+
+    public String toString(String x) {
+        return getStatusIcon() + " | " + this.description;
+    }
```
jx124: Perhaps this could be named something different like `toListString` rather than overloading the `toString` function since it can be confusing as it differs in implementation only in `Event` and `Deadline`.

src/test/java/moss/StorageTest.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @throws MossException If there's an issue with loading or saving tasks.
+     */
+    @Test
+    public void testLoadAndSaveTasks() throws MossException {
+        // Create sample tasks
+        List<Task> tasks = new ArrayList<>();
+        tasks.add(new ToDo("Buy groceries"));
+        tasks.add(new Deadline("Finish homework", LocalDate.parse("2023-09-15")));
+        tasks.add(new Event("Party", LocalDate.parse("2023-09-20"), LocalDate.parse("2023-09-22")));
```
jx124, line 30: I like this event!

src/main/java/Moss/UI.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * The UI class handles user interface interactions and provides messages for the application.
+ */
+public class UI {
+
+    /**
+     * Displays a greeting message when the application starts.
+     */
+    public void greeting() {
```
jx124: Should start with a verb as well. Perhaps `showGreeting`?

src/test/java/moss/MossTest.java
```diff
@@ -0,0 +1,17 @@
+package moss;  //same package as the class being tested
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MossTest {
```
jx124, line 7: Not really sure if these tests are needed?

jx124 (commented):
Great attempt!

PR #438, opened 2023-08-30
[dinde2004] iP

dinde2004:
Thinh's chatbot frees your mind of having to remember things to do. It's

- text-based
- easy to use
- SUPER FAST and ACCURATE response

All you need to do is:

1. download the .jar file from [here](https://github.com/dinde2004/ip/releases/tag/A-Jar).
2. execute the .jar file by using command `java -jar path/to/jar/file`.
3. add your tasks.
4. let the chatbot handle tasks for you 😉.

And the most important thing, of course, it's **FREE**!

Features:

- [x] Manage many types of tasks: todo, deadline, event.
- [x] Find the tasks with matching keyword.
- [x] And many more interesting features coming soon!

If you are a Java programmer, you can download the project from [here](https://github.com/dinde2004/ip). Here is the `main` method:

```java
public class Main {
        Application.launch(Duke.class, args);
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+                dataFile.createNewFile();
+                System.out.println("    Created new data file");
+            } catch (IOException ioe) {
+                System.out.println("    " + ioe.getMessage());
+            }
+        }
+    }
+
+    public void run() {
+        ui.hello();
```
seewhyjay: Might be more useful to name this method `showHelloMessage()`

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+            String description = task.substring(6, slash1 - 1);
+            String from = task.substring(slash1 + 6, slash2 - 1);
+            String to = task.substring(slash2 + 4);
+            Event event = new Event(description, from, to);
+            taskList.addTask(event);
+            ui.addTask(event, taskList.size());
+        } else {
+            throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
seewhyjay: Might be more intuitive to the reader if you catch the invalid command within `processCommand` rather than `addTask`

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,38 @@
...
+    public void printTaskList() {
+        int index = 0;
+        for (Task task: this.taskList) {
+            System.out.println("    " + (++index) + "." + task.toString());
+        }
+    }
+
+    public int size() {
+        return this.taskList.size();
+    }
```
seewhyjay: I'm not sure whether this is counted as a getter, but if it is, maybe `getSize` may be a more appropriate method name

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class Ui {
+    private final String HORIZONTAL_LINE = "    _______________________________________________________________";
+    public void printList(TaskList taskList) {
+        System.out.println("    Here are the tasks in your list:");
+        taskList.printTaskList();
+    }
+
+    public void hello() {
+        System.out.println(HORIZONTAL_LINE);
```
seewhyjay: Since you have a `printHorizontalLine()` method, maybe you can use that instead.

seewhyjay (commented):
Great PR! Everything is really easy to understand.

src/main/java/duke/data/Storage.java
```diff
@@ -0,0 +1,77 @@
+package duke.data;
+
+import duke.task.*;
```
limjunxian1: Imported classes should always be listed explicitly even though in this case, your Storage class utilizes all the classes in your task package as it gives an excellent documentation value for the class at hand and makes the class easier to comprehend and maintain.

src/main/java/duke/data/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class Storage {
```
limjunxian1, line 18: Write descriptive header comments for your Storage class, which I have also otherwise noticed that all your public classes lack descriptive header comments.

src/main/java/duke/data/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+    private String dataPath;
+
+    public Storage(String dataPath) {
+        this.dataPath = dataPath;
+    }
+
+    public void writeFile(TaskList taskList) throws IOException {
```
limjunxian1, line 31: Write descriptive header comments for your writeFile method, which I have also otherwise noticed that all your public methods lack descriptive header comments.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Override
+    public String writeFile() {
+        return "D | " + super.writeFile() + " | " + this.by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " +
```
limjunxian1: I like how you adhere to the coding standard that any line length should be no longer than 120 chars by practicing proper indentation, which improved readability.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
+package duke;
```
limjunxian1, line 1: I like how the the ordering of import statements in all of your classes are consistent, making it easier to browse the list and determine the dependencies

limjunxian1 (commented):
Overall, I found your code easy to read because of your proper use of indentation and Egyptian braces. However, I noted a few minor coding standard violations too which I have highlighted in my comments. Otherwise, LGTM!

PR #446, opened 2023-08-31
[waseemingly] iP

waseemingly:
# Duke :computer:
“We always need little reminders that it's gonna be all good.” – Andy Grammer ([source](https://www.brainyquote.com/quotes/andy_grammer_870566?src=t_reminders))

Duke is your daily task reminder. It's,
- text-based
- ~~EASY~~ _SUPER_ EASY to learn
- quick to use

All you need to do is, 
1. download it from [here](https://github.com/nus-cs2103-AY2324S1/ip/pull/446).
2. add your tasks, deadlines, events
3. allow it to manage it 😉

And it is **FREE!**

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events 
- [ ] Reminders (coming soon)

```
public void run() {

        while (true) {
            try {
                String userInput = scanner.nextLine();
                Command command = Parser.parse(userInput); // This interprets the user input and returns the respective command

                command.execute(tasks, ui, storage); // Execute the command

                if (command instanceof ByeCommand) {
                    break;
                }
            } catch (DukeException e) {
                ui.showError(e.getMessage());
            } catch (IOException e) {
                ui.showSavingError(e.getMessage());
            }
        }
    }

public static void main(String[] args) throws FileNotFoundException {
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Constructs a Duke object and initializes the UI, Storage, and TaskList.
+     *
+     * @param filePath The path to the file where tasks are saved and loaded.
+     * @throws FileNotFoundException If the file is not found.
+     */
+    public Duke(String filePath) throws FileNotFoundException {
+        ui = new Ui();
+        ui.showWelcome();
```
cheeggered: Any reason why you put this line in the constructor instead of the run() method body?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Parses the given user input string into a Command object.
+     * This method recognizes various command keywords and constructs the appropriate command object.
+     *
+     * @param userInput The raw input string from the user.
+     * @return The appropriate Command object based on the user input.
+     * @throws DukeException If the user input is not recognized or improperly formatted.
+     */
+    public static Command parse(String userInput) throws DukeException {
+        if ("bye".equalsIgnoreCase(userInput)) {
```
cheeggered, line 31: good use of equalsIgnoreCase method!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,147 @@
...
+        if (!file.exists()) {
+            return tasks; // return empty list if file doesn't exist
+        }
+
+        try (Scanner scanner = new Scanner(file)) {
+            while (scanner.hasNext()) {
+                String line = scanner.nextLine();
+                String[] parts = line.split(" \\| ");
+
+                switch (parts[0]) {
```
cheeggered: Good use of switch for neater code!

gradle/wrapper/gradle-wrapper.jar
MadLamprey, line 1: Since this is a binary file, maybe could ignore this from tracking?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,77 @@
...
+            return new ByeCommand();
+        } else if ("list".equalsIgnoreCase(userInput)) {
```
MadLamprey, line 33: Could make use of switch to eliminate multiple "else if"

MadLamprey (commented):
LGTM!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,66 @@
...
+        storage = new Storage(filePath);
+        tasks = new TaskList(storage.loadTasks());
+    }
```
H1410101: It seems to me that while well-handled, the bot does not start if the file has not already been created (and the only way out might be to manually create a starting file). Is this a deliberate design decision?

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,15 @@
+package duke;
+
+/**
+ * Represents exceptions specific to the Duke application.
+ */
+public class DukeException extends Exception {
+    /**
+     * Constructs a new DukeException with the specified detail message.
+     *
+     * @param message The detail message. The detail message is saved for later retrieval by the {@link #getMessage()} method.
```
H1410101: I love this `@link`. Just thought I'd mention it.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,77 @@
...
+            } catch (NumberFormatException e) {
+                throw new DukeException("Please provide a valid task number.");
+            }
+        } else if (userInput.startsWith("find")) {
+            String keyword = userInput.substring(4).trim();
+            return new FindCommand(keyword);
+        } else {
+            throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
+    }
```
H1410101: To add on to the two previous reviewers' comments, and tie them together:

I like the fact that you ignore case. However, I strongly believe a switch statement here would be far easier to read.
Perhaps you could use `switch(userInput.toLowerCase())`?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,142 @@
...
+    public Task get(int index) {
+        return tasks.get(index);
+    }
+
+    /**
+     * Deletes a task from the list.
+     *
+     * @param index The index of the task to be deleted.
+     */
+    public void delete(int index) {
```
H1410101, line 53: Nit: perhaps return the Task? seems like behaviour that makes sense, even though I do think IntelliJ will draw a squiggly.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,142 @@
...
+                System.out.println("____________________________________________________________");
+                System.out.println("OK, I've marked this task as not done yet:");
+            }
+            System.out.println("  [" + task.getStatusIcon() + "] " + task.description);
+            System.out.println("____________________________________________________________");
+        } catch (NumberFormatException e) {
+            System.out.println("Please provide a valid task number.");
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("That task doesn't exist!");
+        }
```
H1410101: While it's good that this handles its own exceptions, I do think that UI should be handled by `Ui.java` for maximal OOP-ness; while it's troublesome, I hope it's a good time to do so regardless since there's a shift to GUI anyway.

H1410101 (approved):
LGTM
I like your choice of method names!
Apart from some functionality/design nits, I think your code is great.

PR #452, opened 2023-08-31
[HollaG] iP

HollaG:
# ElonMusk
> If Elon Musk fights Zuckerberg, who will win? Answer: The world

ElonMusk frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

--- 
```
public class Duke {
        new Duke(String.valueOf(SAVE_FILE_LOCATION)).run();
    }
}
```


src/main/java/Duke/Exceptions/DukeException.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param message the error message
+     */
+    public DukeException(String message) {
+        super(message);
+    }
+
+    /**
+     * Returns the error message for this exception.
+     *
+     * @return
```
RoeReRe: Perhaps you missed out the return statement for the Javadoc? Also, is there a particular reason why we cannot just call getMessage rather than wrapping it into another getErrorMessage?

src/main/java/Duke/Parser/Parser.java
```diff
@@ -0,0 +1,213 @@
+package Duke.Parser;
+
+import Duke.Exceptions.DukeException;
+import Duke.Storage.Storage;
+import Duke.Tasks.*;
```
RoeReRe: Blanket import statement might not be allowed for iP.

src/main/java/Duke/Parser/Parser.java
```diff
@@ -0,0 +1,213 @@
...
+public class Parser {
+    private static String UNKNOWN_COMMAND = "\uD83D\uDE21 This command is not something I can handle!";
+
+    private static String NAME_EMPTY = "\uD83D\uDE21 Your item name cannot be empty!";
+    private static String DEADLINE_EMPTY = "\uD83D\uDE21 Missing deadline!";
+
+    private static String FROM_EMPTY = "\uD83D\uDE21 Missing from!";
+    private static String TO_EMPTY = "\uD83D\uDE21 Missing to!";
+    private static String TIME_FORMAT_ERROR = "\uD83D\uDE21 Time format invalid!";
+    private static String INVALID_DATE_FORMAT = "\uD83D\uDE21 Invalid date format! Try using YYYY-MM-DD";
```
RoeReRe: I really like your organisation of the error messages here but perhaps we could put them under an enum for greater clarity?

src/main/java/Duke/Storage/Storage.java
```diff
@@ -0,0 +1,172 @@
+package Duke.Storage;
+
+import Duke.Exceptions.DukeException;
+import Duke.Tasks.*;
```
RoeReRe: Same thing with the blanket import.

src/main/java/Duke/Parser/Parser.java
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @param inputString The input string entered by the user.
+     * @param taskList    The list of tasks
+     * @param storage     The storage object that handles saving updates.
+     * @return true if the program can continue, false if the program has to halt.
+     * @throws DukeException
+     */
+    public static boolean parse(String inputString, TaskList taskList, Storage storage) throws DukeException {
+        String inputCommandString = (inputString.split(" ")[0].toUpperCase());
+        if (!Commands.contains(inputCommandString)) {
```
RoeReRe: We could, actually remove the need for the contains static method if we include an INVALID_COMMAND value in the Commands enum, and have the switch statement below throw the exception when inputCommand is INVALID_COMMAND.

RoeReRe (commented):
Hi! I was assigned to you for the peer PR review for the iP.

I really like the way you implemented and designed the bot. It was very easy to piece out and see how different parts worked together. As this round focuses on the coding standards, I think generally its good. But I do see some inconsistency with blank lines in your code.

All the best!

src/main/java/Duke/Parser/Parser.java
DistractedCat: Looks like this could be better used in an enum.

src/main/java/Duke/UI/Ui.java
```diff
@@ -0,0 +1,126 @@
...
+        while (true) {
+            try {
+                inputString = keyboard.nextLine();
+                printDivider();
+                boolean canContinue = Parser.parse(inputString, taskList, storage);
+                if (!canContinue) {
+                    break;
+                }
+                printDivider();
+            } catch (Exception e) {
```
DistractedCat: It might be better to avoid pokemon exceptions and catch specific ones for better debugging.

src/main/java/Duke/UI/Ui.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Prints feedback to the user on what and how a Task got modified,
+     * based on the user's command.
+     *
+     * @param command
+     */
+    public static void printResult(Commands command, Optional<Task> task, TaskList taskList) {
+        switch (command) {
```
DistractedCat: Good use of the switch statement.

DistractedCat (commented):
LGTM!

PR #454, opened 2023-08-31
[p-xp] iP

p-xp:
# Remy
> "The hand that mocked them, and the heart that fed" -- From Ozymandias by Percy Bysshe Shelley ([source](https://www.poetryfoundation.org/poems/46565/ozymandias))

Ever wanted to get insulted by a task manager that doesn't get much done at all? Remy is here to fill that niche.

What Remy does:
- _Add_ and _remove_ tasks from your tasklist
- Mark tasks as completed / not completed ✅❌
- [ ] Provides a good user experience

Want to get mocked? Here's the fastest way for that:
1. Run the application
2. Type a command that does not exist

---
You can launch Remy from the `main` method:
```
public class Main {
    }
}
```


src/main/java/Remy/ChatbotException.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    /**
+     * Returns String representation of the Exception.
+     * @return String representation of the Exception.
+     */
+    @Override
+    public String toString() {
+        return this.getMessage();
+    }
```
euchangxian: I noticed that the `toString()` method in `ChatbotException` has been overridden to return the error message. The default behaviour of the `Exception::toString` method already includes this message. Unless you meant to provide a custom behaviour, you should consider removing this overridden method to simplify the code!

src/main/java/Remy/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+import remy.command.TodoCommand;
+import remy.command.UnmarkCommand;
+
+/**
+ * A class that makes sense of user inputs.
+ */
+public class Parser {
+
+    /**
+     * Parses user input and carries out any follow-up action (e.g. remy.remy.Task.remy.remy.Task creation).
```
euchangxian, line 23: I believe the extra `remy.`s in `remy.remy.Task` could be an unintentional mistake! Do consider reverting this change.

src/main/java/Remy/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+        case EventCommand.COMMAND_WORD:
+            return new EventCommand(input);
+        case DeleteCommand.COMMAND_WORD:
+            return new DeleteCommand(input);
+        case FindCommand.COMMAND_WORD:
+            return new FindCommand(input);
+        case ExitCommand.COMMAND_WORD:
+            return new ExitCommand();
+        default:
+            throw new ChatbotException("No such command lah.");
```
euchangxian, line 52: Nice use of the `default` branch! 👍

src/main/java/Remy/Remy.java
```diff
@@ -0,0 +1,60 @@
...
+    private TaskList tasks;
+    private Ui ui;
+
+
+    /**
+     * Sets up a Remy chatbot for use.
+     * @param filePath location to store tasklist after ending each session.
+     */
+    public Remy(Path filePath) {
+        ui = new Ui();
```
euchangxian: I noticed `ui` is instantiated, but no instance methods of `ui` is invoked. Perhaps you might want to change the existing static methods to instance method. This will allow for easier unit testing of `Ui` as well, since static methods cannot be overridden to create stubs.

src/main/java/Remy/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+import remy.task.TaskList;
+
+import java.util.Scanner;
+
+public class Ui {
+
+    public final Scanner scanner;
+    private static String divider = "____________________________________________________________\n";
+    private static String shortDivider = "_____________\n";
+    private static String welcomeMessage =
```
euchangxian: You could prefix `final` to these variables that I assume are meant to be constants? If so, do remember to change their names to `DIVIDER`, `SHORT_DIVIDER` and `WELCOME_MESSAGE` respectively to align with the coding standards.

src/main/java/Remy/Task/TaskList.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Returns String representation of all TaskList Tasks that match the given searchString.
+     * @param searchString To compare against Tasks in TaskList.
+     */
+    public String findMatchingTasks(String searchString) {
+        String matchingTasksString = "";
+        for (Task task : this.tasks) {
+            if (task.getDescription().toLowerCase().contains(searchString)) {
```
euchangxian, line 71: Good work converting the strings to lower case for case-insensitive comparison!👍

src/main/java/Remy/command/DeadlineCommand.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param storage Handles saving the updated TaskList.
+     * @throws ChatbotException if the time information is in the wrong format.
+     */
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws ChatbotException {
+        try {
+            Deadline temp = new Deadline(this.taskName, this.dueDate);
+            taskList.add(temp);
+            storage.save(taskList);
+            Ui.printAddedTask(temp, taskList.size());
```
euchangxian: Consider changing `Ui::printAddedTask` to an instance method instead!

src/main/java/Remy/command/Command.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * A representation of a Command that parses user input and follows-up with the relevant action.
+ */
+public abstract class Command {
+
+    /**
+     * Whether the Task should trigger the end of the current Chatbot session.
+     * @return boolean indicating Exit status.
+     */
+    public abstract boolean isExit();
```
euchangxian: This could possibly be a concrete method that returns `false` by default, and only overridden by `ExitCommand` to return `true`. This is so that not every subclass that extends `Command` have to override it, improving simplicity of your code.

text-ui-test/EXPECTED.TXT
```diff
@@ -1,7 +1,29 @@
-Hello from
- ____        _        
-|  _ \ _   _| | _____ 
-| | | | | | | |/ / _ \
-| |_| | |_| |   <  __/
-|____/ \__,_|_|\_\___|
+____________________________________________________________
+
+I'm Remy.Remy, and it is NOT nice to see you.
```
euchangxian, line 3: Possibly an oversight here `Remy.Remy`. Consider reverting this change!

text-ui-test/runtest.bat
```diff
@@ -15,7 +15,7 @@ IF ERRORLEVEL 1 (
 REM no error here, errorlevel == 0
 
 REM run the program, feed commands from input.txt file and redirect the output to the ACTUAL.TXT
-java -classpath ..\bin Duke < input.txt > ACTUAL.TXT
+java -classpath ..\bin remy.remy < input.txt > ACTUAL.TXT
```
euchangxian, line 18: Same issue as above!

euchangxian (commented):
Overall LGTM. Great implementation of the Commands class!

src/main/java/Remy/command/FindCommand.java
```diff
@@ -0,0 +1,42 @@
...
+
+    // Regex pattern for "find {keyword}" format
+    private static final Pattern FIND_PATTERN = Pattern.compile("^find\\s+(\\S+)$", Pattern.CASE_INSENSITIVE);
+
+    public FindCommand(String input) throws ChatbotException {
+        Matcher matcher = FIND_PATTERN.matcher(input.trim());
+        if (matcher.matches()) {
+            this.searchString = matcher.group(1).toLowerCase();
+        } else {
+            throw new ChatbotException("format wrong bro");
```
remuslum: Perhaps on top of telling "bro" that his format is wrong, you could additionally provide the correct format to him so "bro" knows what he is doing wrong. E.g find ___ -> Did you forget to input a word?

src/test/java/Remy/command/DeadlineCommandTest.java
```diff
@@ -0,0 +1,40 @@
...
+import Remy.ChatbotException;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class DeadlineCommandTest {
+    @Test
+    public void DeadLineCommand_missingInfo_throwChatbotException() {
+        assertThrows(ChatbotException.class, () -> new DeadlineCommand("deadline"));
```
remuslum, line 14: I like the way you named your function to include the type of error like missing info and invalid date. To make it even better, might I suggest that instead of naming 1,2 and making it general, perhaps a more specific title could be used e.g invalid_date_missing_month, invalid_date_missing_year

src/test/java/Remy/command/MarkCommandTest.java
```diff
@@ -0,0 +1,44 @@
...
+        assertThrows(ChatbotException.class, () -> new MarkCommand("mark"));
+    }
+
+    @Test
+    public void MarkCommand_invalidIndex_throwChatbotException() {
+        assertThrows(ChatbotException.class, () -> new MarkCommand("mark abc"));
+    }
+
+    @Test
+    public void MarkCommand_negativeIndex_throwChatbotException() {
```
remuslum: Great job here. The titles here are not too general and can be easily comprehended 👍🏻

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,26 @@
...
+            this.index = index;
+        }
+    }
+
+    @Override
+    public boolean isExit() {
+        return false;
+    }
+
+    @Override
```
remuslum: I don't think there is a need to provide an overwrite tag for the execute method as the method has not been implemented in the Command class.

PR #455, opened 2023-08-31
[kenvynKwek] iP

kenvynKwek:
# DogeBot :dog2:
> "Im your best friend that helps you manage your tasks c:" -DogeBot

DogeBot remembers all the things you need to do with its **excellent** memory, DogeBot is:
- a dog
- ~~cute~~friendly
- _easy_ to use

All you need to do is:
1. access it from [here](https://github.com/kenvynKwek/ip)
2. run it
3. add your tasks and take your mind off things :)

Features:
- [x] Manage different types of tasks
- [x] Search keywords in tasks
- [ ] Graphical user interface (coming soon)

Just look at how ~~cute~~friendly DogeBot's response is when you `mark` a task:
```ruby
public static void mark(int index) throws DogeBotException {
        if (tasks.size() == 0) {
            throw new DogeBotException("Oops ! Try adding some tasks first c:");
        }

        tasks.get(index).markTask(true);
        System.out.println("Good job on completing a task ! You deserve a cookie C:");
        System.out.println("\t" + tasks.get(index).toString());
    }
```


src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Event(String description, String start, String end, boolean isDone) {
+        super(description, isDone);
+        this.start = new DateTimeHandler(start);
+        this.end = new DateTimeHandler(end);
+    }
+
+    @Override
+    public String toString() {
+        return "E | " + super.toString() + " | " + start.toString() + " - " + end.toString();
```
mfjkri, line 13: Consider using String.format?

src/main/java/Storage.java
```diff
@@ -0,0 +1,64 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+
+import java.io.FileWriter;
+import java.io.BufferedWriter;
+import java.io.IOException;
```
mfjkri: Could these import statements perhaps be reordered?

src/main/java/Parser.java
```diff
@@ -0,0 +1,56 @@
+import java.util.InputMismatchException;
+import java.util.Scanner;
+
+public class Parser {
+    public boolean scan() {
+        Scanner sc = new Scanner(System.in);
+        boolean isLoop = true;
+
+        try {
+            switch (sc.next().toLowerCase()) {
```
mfjkri: I like the use of a switch statement here which makes the code readable and easy to follow!

src/main/java/DateTimeHandler.java
```diff
@@ -0,0 +1,31 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class DateTimeHandler {
```
mfjkri: Consider writing descriptive header comments for all public classes/methods so that other developers will not be forced to read the code of the method to understand its exact behavior.

mfjkri (commented):
Overall clean and easy to read code. 

Do consider adding javadocs for all the public methods and classes.

src/main/java/dogebot/DogeBot.java
```diff
@@ -0,0 +1,45 @@
...
+            System.out.println(e.getMessage());
+        }
+
+        ui = new Ui();
+        userInput = new Parser();
+        storage = new Storage(file);
+        tasks = new TaskList(storage.readFromTxtFile());
+    }
+
+    public void run() {
```
mfjkri: Consider adding javadocs here so that readers can easily understand the functionality of the method.

src/main/java/dogebot/DateTimeHandler.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Converts MMM format month to integer.
+     *
+     * @param s MMM format month.
+     * @return month as integer.
+     */
+    public int monthToInt(String s) {
+        int month;
+        switch (s) {
```
elaineshijie, line 89: Switch and case statements have the same indentation, good job!

src/main/java/dogebot/TaskList.java
```diff
@@ -0,0 +1,207 @@
...
+     * @param index Index of task to mark.
+     * @throws DogeBotException If tasklist is empty.
+     */
+    public static void mark(int index) throws DogeBotException {
+        if (tasks.size() == 0) {
+            throw new DogeBotException("Oops ! Try adding some tasks first c:");
+        }
+
+        tasks.get(index).markTask(true);
+        System.out.println("Good job on completing a task ! You deserve a cookie C:");
```
elaineshijie: Would it be better to have methods in Ui to print these messages?

src/main/java/dogebot/TaskList.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Retrieves tasks with the input keyword.
+     *
+     * @param s Input keyword.
+     * @throws DogeBotException If input keyword is empty.
+     */
+    public static void find(String s) throws DogeBotException {
+        if (s.isBlank()) {
+            throw new DogeBotException("Oops ! The keyword cannot be empty :(");
```
elaineshijie, line 241: Would it be better if exceptions are created based on the events eg. KeywordNotFoundException and these specific exceptions extends from DogeBotException?

elaineshijie (commented):
Code is generally well done but more can be done to organise code (Ui has methods to print output to user instead as Ui is a class to interact with user).

PR #460, opened 2023-08-31
[Tanveer Singh] iP

tanveersingh10:
# Bot Cristiano

Cristiano provides full support for task management. It is:
- easy to learn
- super convenient 
- interactive user interface 
- Features the great Cristiano Ronaldo since Messi wasn't available

## All you need to do is,
Install from [here](https://github.com/tanveersingh10/ip/releases/tag/A-Jar)
Add your todos / deadlines / events
Let it manage your tasks for you 🔥💯:+1: 

Features:
- [x] managing todos
- [x] managing events and deadlines 
- [x] saving all your important information
- [x] fun user experience
- [ ] reminders: in the future!

```
public class Main {
    }
}
```


src/main/java/duke/main/UI.java
```diff
@@ -0,0 +1,73 @@
...
+        printLine();
+        System.out.println("Hello! I'm " + chatBotName + "! SUIIII!!!");
+        System.out.println("What can I do for you?");
+    }
+
+    public void displayList(ArrayList<Task> list, int numOfTasks) {
+        printLine();
+        System.out.println("Here are the duke.tasks in your list:");
+        for (int i = 0; i < numOfTasks; i++) {
+            System.out.println((i + 1) + ". " + list.get(i));
```
Seonlo99: I think it might be better to get the number of tasks by list.size() instead of passing in an additional parameter.

src/main/java/duke/main/UI.java
```diff
@@ -0,0 +1,73 @@
...
+        printLine();
+    }
+
+    public void exit() {
+        printLine();
+        System.out.println("Bye. Hope to see you again soon!");
+        printLine();
+    }
+    public void addTask(String taskName, int numOfTasks) {
+        printLine();
```
Seonlo99: I think it will be better if there is a print message function that includes the print line so that printLine() will  not be needed at every command.

src/main/java/duke/exceptions/DukeException.java
```diff
@@ -0,0 +1,11 @@
+package duke.exceptions;
```
Seonlo99, line 1: I think it's really great of you to implement a variety of Exception classes.

src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+                    input = scanner.nextLine();
+                    String command = this.parser.getCommand(input);
+                    switch (command) {
+                        case "list":
+                            taskManager.list();
+                            break;
+                        case "mark":
+                        case "unmark":
+                            handleMarking(input, taskManager);
+                            break;
```
Seonlo99: I think you might have forgotten to handle the "mark" case.

src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+            } catch (NumberFormatException e) {
+                throw new InvalidArgumentException("Please enter a numerical index!");
+            } catch (StorageException e) {
+                throw new StorageException("duke.main.Storage Error");
+            }
+        }
+
+        private void updateStorage() throws StorageException {
+            this.storage.saveData(taskManager);
+        }
```
Seonlo99: I think it might be better if all the command handling are contained inside the TaskManager class.

src/main/java/duke/main/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+import duke.tasks.Task;
+import duke.tasks.ToDo;
+import duke.tasks.Event;
+import duke.tasks.Deadline;
+
+import duke.exceptions.StorageException;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.io.*;
```
Seonlo99, line 13: I think it will be better if you just import the classes you need instead of the whole package.

Seonlo99 (commented):
LGTM. just some nits to fix.

src/main/java/duke/main/TaskManager.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Marks a task as done by its index in the list and updates the UI.
+     *
+     * @param index The index of the task to be marked.
+     * @throws InvalidArgumentException If the index is out of range.
+     */
+    public void mark(int index) throws InvalidArgumentException {
+        if (index > numOfTasks) {
+            throw new InvalidArgumentException("I'm sorry but that task does not exist. There are only " + numOfTasks + "duke.tasks.");
```
dinde2004, line 72: I think that one line should not contain more than 120 characters. You can consider line wrapping.

src/main/java/duke/main/TaskManager.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Marks a task as not done by its index in the list and updates the UI.
+     *
+     * @param index The index of the task to be unmarked.
+     * @throws InvalidArgumentException If the index is out of range.
+     */
+    public void unmark(int index) throws InvalidArgumentException {
+        if (index > numOfTasks) {
+            throw new InvalidArgumentException("I'm sorry but that task does not exist. There are only " + numOfTasks + " duke.tasks.");
```
dinde2004, line 90: I think that one line should not contain more than 120 characters. You can consider line wrapping.

src/main/java/duke/main/TaskManager.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Removes a task from the ArrayList of tasks.
+     *
+     * @param index The index of the task to be unmarked.
+     * @throws InvalidArgumentException If the index is out of range.
+     */
+    public void delete(int index) throws InvalidArgumentException {
+        if (index > numOfTasks) {
+            throw new InvalidArgumentException("I'm sorry but that task does not exist. There are only " + numOfTasks + " duke.tasks.");
```
dinde2004, line 108: I think that one line should not contain more than 120 characters. You can consider line wrapping.

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,83 @@
+package duke.tasks;
+
+/**
+ * The Task class represents a basic task with a name and completion status.
+ * It provides methods to work with tasks, such as marking them as done and retrieving their details.
+ */
+public class Task {
+    protected String taskName;
+    protected boolean done;
```
dinde2004: I think you should name boolean sound like boolean, for example: isDone instead of done.

dinde2004 (commented):
Overall, despite some coding standard violations like naming and length of line, your code is very good. I really like the way you use a wide variety of exceptions.

PR #469, opened 2023-08-31
[ong-wei-hong] iP

ong-wei-hong:
# GOAT
GOAT stands for
 - greatest
 - of 
 - all
 - time
GOAT has the following commands
1. list
2. todo
3. deadline
4. event
5. mark
6. unmark
7. delete
8. find

A GOAT bot can be initialised as below
```java
public class Main {
                Duke duke = new Duke();
        }
}
```
We have
 - [X] Todos
 - [X] Deadlines
 - [X] Events
 - [ ] Dragons? :dragon:
 > And then there were Dragons - someone [source](https://www.goodreads.com/series/332001-and-they-found-dragons)

Initialise `Duke` **today**!


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+
+public class Duke {
+
+  private final String BOT_NAME = "GOAT";
+  private final String FILEPATH = "data/";
+  private final String SAVEFILE_NAME = "save.txt";
```
StevenLiudw: Consistent naming conventions are followed. However, it's better to follow the Java naming convention of having constant variables (BOT_NAME, FILEPATH, etc.) as static final

src/main/java/duke/command/AddTaskCommand.java
StevenLiudw, line 1: While the code is relatively clear, it could benefit from more in-line comments or JavaDoc comments, especially for the public methods and classes.

src/main/java/duke/helper/DatetimeHelper.java
```diff
@@ -0,0 +1,22 @@
+package duke.helper;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class DatetimeHelper {
+
+  public static String inputFormat = "yyyy-MM-dd HHmm";
+  public static String exampleDatetime = "2015-09-15 1300";
```
StevenLiudw: inputFormat and exampleDatetime should be made final constants since they are not meant to change.

src/main/java/duke/io/FileIO.java
```diff
@@ -0,0 +1,50 @@
...
+  public String read() throws FileNotFoundException {
+    File f = new File(filepath + filename);
+    Scanner in = new Scanner(f);
+    StringBuilder sb = new StringBuilder();
+    while (in.hasNext()) {
+      sb.append(in.nextLine().trim());
+      sb.append('\n');
+    }
+    return sb.toString().trim();
+  }
```
StevenLiudw: inputFormat and exampleDatetime should be made final constants since they are not meant to change.

src/main/java/duke/io/Printer.java
```diff
@@ -0,0 +1,60 @@
...
+        (s) -> {
+          System.out.println(String.format("    %s", s));
+        },
+        () -> {
+          messages.forEach(
+              (s) -> {
+                System.out.printf("    %s\n", s);
+              });
+        });
+    System.out.println(line);
```
StevenLiudw: Rather than checking if the error is present using ifPresentOrElse, it might be clearer to check error.isPresent() and then act accordingly.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,60 @@
...
+      throw new DukeException("The from of a event cannot be empty.");
+    }
+    try {
+      this.from = DatetimeHelper.parse(from);
+    } catch (DateTimeParseException e) {
+      throw new InvalidDatetimeFormatException("from", "event");
+    }
+
+    if (to.isEmpty()) {
+      throw new DukeException("The to of a event cannot be empty.");
```
StevenLiudw: Instead of "a event", use "an event".

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  public void addTask(Task t) {
+    tasks.add(t);
+  }
+
+  public String getNumberOfTasks() {
+    return String.format(
+        "Now you have %d task%s in the list.", tasks.size(), tasks.size() == 1 ? "" : "s");
```
StevenLiudw: For clarity and consistency, consider renaming the method getNumberOfTasks() to something like getTaskCountMessage().

StevenLiudw (commented):
LGTM. Quite impressed by the nice OOP structure(divide exception, helper classes and IO). But it is clear that you need to provide more JavaDoc for better user understanding. :)

src/main/java/duke/Duke.java
glenngnng: Should FILEPATH be changed to FILE_PATH? Perhaps can use underlines to separate the words in constant naming. Same for SAVE_FILE_NAME 😄

src/main/java/duke/command/AddTaskCommand.java
glenngnng: Agreed. Javadocs here would be useful.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,61 @@
...
+  public static final String DEADLINE = "deadline";
+  public static final String EVENT = "event";
+  public static final String DELETE = "delete";
+	public static final String FIND = "find";
+
+  protected Printer out;
+  protected TaskList taskList;
+  private FileIO savefile;
+
+  private boolean saveAfterAction = false;
```
glenngnng: Perhaps it would be better to change the name to include a prefix like is? A possible change could be isSavedAfterAction. This could be easier to understand that it is a boolean.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+  public DeleteCommand(Printer out, TaskList taskList, FileIO savefile, String name) {
+    super(out, taskList, savefile);
+    this.s = name;
+  }
+
+  @Override
+  public void action() {
+    Task task;
+    try {
+      int i = Integer.parseInt(s); // 1-indexed
```
glenngnng: Not sure if this is OK under the Java coding standard, but perhaps can indent the comment under this line.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public String toCommand() {
+    return String.format("%s /by %s", super.toCommand(), DatetimeHelper.commandFormat(by));
+  }
+
+  @Override
+  public boolean filter(Optional<LocalDateTime> before) {
+    return before.filter((datetime) -> datetime.isBefore(by)).isEmpty(); // if is empty return true
```
glenngnng: Same issue with commenting here. Perhaps can push the comment to the next line and indent it the same way. Improves readability if the line is too long.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,60 @@
...
+        "%s /from %s /to %s",
+        super.toCommand(), DatetimeHelper.commandFormat(from), DatetimeHelper.commandFormat(to));
+  }
+
+  @Override
+  public boolean filter(Optional<LocalDateTime> before) {
+    return before
+        .filter((datetime) -> datetime.isBefore(from))
+        .isEmpty(); // if is empty return true
+  }
```
glenngnng: Perhaps some javadocs explaining these methods would be useful?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,69 @@
...
+
+    if (description.isEmpty()) {
+      throw new DukeException(
+          String.format(
+              "The description of a %s cannot be empty.",
+              this.getClass().getSimpleName().toLowerCase()));
+    }
+  }
+
+  public static Task createTask(String type, String name, Map<String, String> arguments) {
```
glenngnng: I like this naming, it is very clear! 👍

PR #485, opened 2023-09-01
[Joey Lee] iP

applepiofmyeye:
###Barbie Chatbot

> "Humans only have one ending. Ideas live forever." -- Ruth Handler ([source](https://collider.com/barbie-best-quotes-ranked/#quot-humans-only-have-one-ending-ideas-live-forever-quot)

Let your ideas live forever with this barbie chatbot. It's,

- Text-based
- ~FAST~ _SUPER_ FAST
- And Barbie approved!

All you need to do is, 
1. Download it 
2. Double click it
3. Add your tasks
4. Let it manage your tasks for you 😄 

And it's **FREE**!

Features:
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Find tasks
- [ ] List tasks

If you're a Java user, you can use it to practise Java too! Here's the `main` method

```
public class Main {
    }
}
```


src/main/java/Barbie.java
```diff
@@ -0,0 +1,182 @@
+import types.Deadlines;
+import types.Party;
+import types.Task;
+import types.Todo;
+import exceptions.*;
```
javinchua: It would be better to avoid wildcard imports, rather you could list out the imports of the classes that are being used.

src/main/java/Barbie.java
```diff
@@ -0,0 +1,182 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.*;
+
+
+public class Barbie {
+
+    enum Command {
+        MARK,
+        UNMARK,
+        DEL,
```
javinchua: I like that you have created an enumeration for the commands but perhaps it might be better to use the full word naming for the enum value of DELETE instead of DEL.

src/main/java/Barbie.java
```diff
@@ -0,0 +1,182 @@
...
+                            String[] parts2 = parts[1].split("/");
+
+                            switch (command) {
+                                case DEADLINE:
+                                    if (parts2.length < 2) {
+                                        throw new BarbieNoDeadlineException();
+                                    }
+
+                                    desc = parts2[0];
+                                    LocalDate by = LocalDate.parse(parts2[1]);
```
javinchua: Before accessing parts2[1], it would be good to ensure that parts2 actually has 2 elements, to avoid running into ArrayOutOfBoundsException.

src/main/java/Barbie.java
```diff
@@ -0,0 +1,182 @@
...
+            while (true) {
+                try {
+
+                    String input = scanner.nextLine();
+                    String[] parts = input.split(" ", 2);
+                    Command command = Command.valueOf(parts[0].toUpperCase());
+
+                    Ui.barbie();
+
+                    switch (command) {
```
javinchua: The main class, Barbie, seems to be handling many responsibilities. Consider using a command class to break down the responsibilities.

src/main/java/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+                    String taskStatus = taskParts[1];
+                    String desc = taskParts[2];
+                    Task task;
+
+                    if (Objects.equals(taskType, "T")) {
+                        task = new Todo(desc);
+                    } else if (Objects.equals(taskType, "D")) {
+                        task = new Deadlines(desc, LocalDate.parse(taskParts[3]));
+                    } else if (Objects.equals(taskType, "P")) {
+                        task = new Party(desc, LocalDate.parse(taskParts[3]), LocalDate.parse(taskParts[4]));
```
javinchua:
```suggestion
                    if (String.equals(taskType, "T")) {
                        task = new Todo(desc);
                    } else if (String.equals(taskType, "D")) {
                        task = new Deadlines(desc, LocalDate.parse(taskParts[3]));
                    } else if (String.equals(taskType, "P")) {
                        task = new Party(desc, LocalDate.parse(taskParts[3]), LocalDate.parse(taskParts[4]));
```
Consider using string.equals for this scenario as it is more specific to the data types.

javinchua (commented):
Overall, I enjoy the abstraction that you have done to the Duke Class, especially in the Storage class, where you specifically remove particular lines depending on the command. An overall comment could be to add more Javadocs, especially for public methods as it will allow improved overall readability, specifically in the case of when u have overloaded methods, as this will allow the author and others to better understand the different scenarios for using the overloaded methods.

src/main/java/Storage.java
applepiofmyeye: I didn't notice this, thanks!!

src/main/java/Storage.java
```diff
@@ -50,14 +51,15 @@ protected static void addToList(Path path, String desc) {
...
             Files.write(path, line.getBytes(), StandardOpenOption.APPEND);
         } catch (IOException e) {
             System.out.println(e.getMessage());
         }
     }
 
+
     /**
      * Writes to the file given, in the format of [type][status] [desc].
      * Eg. [T][ ] read book
```
miljyy: I like how detailed your Javadoc is, even providing the example of the format that will be written to file.

src/main/java/Barbie.java
```diff
@@ -155,17 +138,32 @@ public static void main(String[] args) {
...
+                    case BYE:
+                        break loop; // break out of the while loop, not switch statement
+
+                    default:
+                        // Editing variables
+                        list.add(indexNumber, new Task(input)); // Create a new Task
+                        indexNumber += 1; //Incrementing item counter
+                        // Output
+                        System.out.println("\t Okey Dokey! I've added this task into your list:\n"
+                                + "\t[ ] " + input);
```
miljyy: Perhaps such output could be handled by Ui class?

miljyy (commented):
Overall I like how clean the code is, generally easy to read and understand!

PR #486, opened 2023-09-01
[peasantbird] iP

peasantbird:
Bert frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/peasantbird/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


lilozz2 (commented):
Was a lot cleaner than my own haha. Had extensive documentation which was something I really appreciate.

Overall code is clean and easy to read especially with the intuitive function names e.g. ensureTaskFileExists();
Comments were given everywhere, which was helpful in understanding the code
Test cases were sufficient and handled a majority of the cases

The only issue that I had was with the while loop and switch statements in the main function which I felt was doing too much and could have been better compartmentalised. The indentation was also going in too deep and starting to become like an arrow head.

src/main/java/Bert.java
```diff
@@ -0,0 +1,354 @@
...
+    private static final String BOT_NAME = "Bert";
+    private static final String FILE_PATH = "./data/bert.txt";
+    private static ArrayList<Task> al = new ArrayList<>();
+
+    /**
+     * Introduces itself.
+     */
+    private static void introduce() {
+        System.out.println(
+                "____________________________________________________________\n" +
```
iyioon: Looks that the underbar line is being reused. Would it be simpler to abstract this such as having a UI do it?

src/main/java/Bert.java
```diff
@@ -0,0 +1,354 @@
...
+        }
+        System.out.println("____________________________________________________________\n");
+    }
+
+    /**
+     * Marks a task at a specific index of the list as done.
+     *
+     * @param index The index of a task on the list
+     */
+    private static void mark(String index) {
```
iyioon: Can you explain the reason behind receiving index as a String?

src/main/java/Bert.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    /**
+     * Marks a task at a specific index of the list as undone.
+     *
+     * @param index The index of a task on the list
+     */
+    private static void unmark(String index) {
+        int i = Integer.parseInt(index) - 1;
+        Task t = al.get(i);
```
iyioon: How about using `Task task = al.get(i)` for the variable name?

src/main/java/Bert.java
```diff
@@ -0,0 +1,354 @@
...
+     * Loads tasks from the save data into the chatbot.
+     */
+    private static void loadTasks() {
+        try {
+            ensureTaskFileExists();
+            File file = new File(FILE_PATH);
+            Scanner sc = new Scanner(file);
+            while (sc.hasNext()) {
+                String task = sc.nextLine();
+                switch (task.charAt(0)) {
```
iyioon: You can follow the [Java Coding Standard](https://se-education.org/guides/conventions/java/intermediate.html#:~:text=The%20switch%20statement%20should%20have%20the%20following%20form%3A) for the switch statement.

iyioon (commented):
LGTM. Some naming can be improved as well as abstractions.

PR #495, opened 2023-09-01
[spatuly] iP

spatuly:
UkePro remembers things you need to do for you. It's,
- text-based
- easy to learn
- ~~FAST~~ _VERY_ FAST to use

All you need to do is,
2. run it
3. add your tasks
4. leave it to remember your tasks for you! 😄 

And it is **FREE**! Get it **NOW**!

Features:
- [x] Managing tasks (with and without deadlines)
- [x] Search for added tasks
- [ ] \(Coming soon) Reminders 

```java
    new Duke(FILEPATH).run();
}
```


src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+        System.out.println(greet);
+    }
+
+    /**
+     * Prints the list of tasks in the TaskList given.
+     *
+     * @param storedTasks TaskList to be printed.
+     * @param isListCommand Whether the method is called from a list command.
+     */
+    public void printList(TaskList storedTasks, boolean isListCommand) {
```
kiwibang: Perhaps can avoid nesting by having a variable to conditionally store different strings depending on the value of isListCommand? 🤔 
For example: 

```java
String initialMessageToUser = isListCommand
        ? "Okay! Here's the full list of your added tasks:"
        :  "Here's the list of matching tasks I found:";
System.out.println(intialMessageToUser);
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+        int len = storedTasks.getLength();
+
+        if (isListCommand) {
+            if (len > 0) {
```
kiwibang: Perhaps can add this as a guard clause at the start of the method? Since the function can terminate if the list is empty 🤔 
For example:

```java
if (len == 0) {
    System.out.println("No tasks found!");
    return; // terminates the method early since there is nothing to print
}
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+                System.out.println("No tasks found!");
+            }
+        } else {
+            if (len > 0) {
+                System.out.println("Here's the list of matching tasks I found:");
+                for (int i = 1; i < len + 1; i++) {
+                    System.out.println(i + ". " + storedTasks.getTask(i - 1));
+                }
+            } else {
+                System.out.println("No matching tasks found :(");
```
kiwibang: Perhaps can add another String variable to store a string value conditionally if this error message needs to be printed out? 🤔 
For example:
```java
String emptyTasksMessage = isListCommand
        ? "No tasks found!"
        : "No matching tasks found :(";
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,271 @@
...
+     */
+    public void run() {
+        try {
+            ui.printGreeting();
+            storage.createDataFile();
+        } catch (Exception e) {
+            ui.printError(e);
+        }
+
+        while (hasNext) {
```
kiwibang: Perhaps this can be shortened to multiple methods? 🤔

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param path File path to the text file which stores task information.
+     */
+    public Duke(String path) {
+        this.ui = new Ui();
+        this.storage = new Storage(path);
+
+        try {
+            tasks = new TaskList(storage.loadTaskList());
+        } catch (DukeException e) {
+            ui.printError(e);
```
kiwibang: Perhaps can avoid using try-catch as a control flow statement? I believe there is a method to check if the database file exists using Java IO classes 🤔

kiwibang (commented):
LGTM! Just some nits to fix. Better than my PR honestly. 👍

src/main/java/duke/DukeDataFileException.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Represents a Duke exception when any error occurs related to the data file.
+ */
+public class DukeDataFileException extends DukeException {
+
+    /**
+     * Returns the string representation of the exception.
+     */
+    @Override
+    public String toString() {
```
thienmy0: Do overriden methods need javadoc? Else is there a return for this method that is not mentioned in the javadoc?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,271 @@
...
+     */
+    public void handleExit() {
+        ui.printExit();
+        hasNext = false;
+    }
+
+    /**
+     * Generates list of added tasks.
+     */
+    public void handleList() {
```
thienmy0: Not sure I like the naming of this method since it doesn't give me the impression that it is meant to generate the list

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @return String representation of the user input.
+     */
+    public String getInput() {
+        return input.nextLine();
+    }
+
+    /**
+     * Greets the user when the chatbot is started.
+     */
```
thienmy0: I like how all your Javadoc comments are very clear and straight to the point!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Represents the Duke chatbot.
+ */
+public class Duke {
+
+    private static final String FILEPATH = "./data/duke.txt";
+    private static boolean hasNext = true;
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
```
thienmy0: Good job on using the appropriate naming convention for variables throughout your code :+1:

thienmy0 (commented):
Your code looks neat to me!
I honestly struggled to find issues.
Keep up the good work :smile:

PR #525, opened 2023-09-03
[butteredyakiimo] iP

butteredyakiimo:
# BUTTER
> "Tracking your tasks will be as smooth as butter. :butter:"

BUTTER helps you remember what you need to do. It's,
- text-based
- easy to use 
- ~FAST~ _SUPER_ FAST to use

All you need to do is, 
1. download it from [here.](https://github.com/butteredyakiimo/ip/releases)
2. double click it.
3. add your tasks.
4. let it manage your tasks for you!

And it is **FREE**!

## Features:
- [x] Managing tasks
- [x] Managing deadlines 
- [x] Managing events
- [x] Searching for tasks
- [x] Reminders (coming soon)

If you are a Java programmer, you can use it to practise Java too. Here's the `main` method: 

```java
        new Duke("./data/duke.txt").run();
    }
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,190 @@
...
+        String[] inputArr = input.split(" ", 2);
+        if (inputArr.length == 1) {
+            throw new NoDescException();
+        } else {
+            if (inputArr[1].isBlank()) {
+                throw new NoDescException();
+            } else {
+                //0 for unmarked, any other number for marked
+                ToDo toDo = new ToDo(0, inputArr[1]);
+                tasks.addTask(toDo, storage);
```
derekjxtan: Perhaps acting on the inputs can be put in another class? IMO the name parser doesn't really indicate that this behaviour is expected.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,121 @@
...
+            if (!dataFile.exists()) {
+                dataFile.createNewFile();
+            }
+        } catch (IOException e) {
+            System.out.println("Unable to create file!!");
+        }
+        return dataFile;
+    }
+
+    public ArrayList<Task> loadTasks() {
```
derekjxtan: personally i feel that Storage should only be responsible for reading what is stored in the file. Maybe the logic for converting the text to the Task could be moved to parser or somewhere else relevant. What do you think?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+            sc.close();
+        } catch (IOException e) {
+            ui.showError(e.getMessage());
+        }
+        return list;
+    }
+
+    //this method should be called by the tasklist class ONLY!!! because we want to
+    //keep the tasklist private
```
derekjxtan: I think rather than adding a comment, it would be better for TaskList to generate a ArrayList of strings for the Storage class to write to file. This ensures that the the ArrayList of Tasks would not be passed to an external class. This make the Storage class more reusable as well.

derekjxtan (commented):
LGTM. Overall great use of OOP principles and code quality is good!

.gitignore
```diff
@@ -15,3 +15,5 @@ bin/
 
 /text-ui-test/ACTUAL.TXT
 text-ui-test/EXPECTED-UNIX.TXT
+
+/src/main/java/*.class
```
ZD292, line 19: Looks good to me 😄

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,50 @@
...
+        Parser parser = new Parser(this.storage, this.tasks, this.ui);
+        Scanner scanner = new Scanner(System.in);
+
+        while (scanner.hasNextLine()) {
+            String input = scanner.nextLine();
+            if (parser.isGoodbye(input)) {
+                break;
+            } else {
+                parser.parseInput(input);
+            }
```
antonTan96: The way you check for a terminating condition is quite intuitive by checking whether the input is "bye" or not. However, this involves throwing the input into the parser twice. Would it be better to let the parser handle the case "bye" in `parser.parseInput(input)` ?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * Returns true if the string can be parsed as an integer.
+     *
+     * @param s the string to be tested.
+     * @return true if s can be parsed as an integer.
+     */
+    //credit: https://www.freecodecamp.org/news/java-string-to-int-how-to-convert-a-string-to-an-integer/
+    private static boolean isNumber(String s) {
+        return s != null && s.matches("[0-9.]+");
+    }
```
antonTan96: I like how regex is used here :D

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,318 @@
...
+            String strIndex = inputArr[1];
+            if (isNumber(strIndex)) {
+                int index = Integer.parseInt(strIndex) - 1; //because index starts from 1
+                if (tasks.isValidTaskId(index)) {
+                    tasks.markTask(index);
+                }
+            } else {
+                throw new InvalidTaskIdException();
+            }
+        }
```
antonTan96: What happens if `tasks.isValidTaskId(index)` returns false? Could InvalidTaskIdException be thrown? I would write this part like this:
```
        if (inputArr.length == 1) {
            throw new NoTaskIdException();
        } else {
            String strIndex = inputArr[1];
            if (isNumber(strIndex) && tasks.isValidTaskId(Integer.parseInt(strIndex) - 1)) {
                int index = Integer.parseInt(strIndex) - 1; //because index starts from 1
                
                tasks.markTask(index);
                
            } else {
                throw new InvalidTaskIdException();
            }
        }
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,318 @@
...
+            String strIndex = inputs[1];
+            if (isNumber(strIndex)) {
+                int index = Integer.parseInt(strIndex) - 1; //because index starts from 1
+                if (tasks.isValidTaskId(index)) {
+                    tasks.unMarkTask(index);
+                }
+            } else {
+                throw new InvalidTaskIdException();
+            }
+        }
```
antonTan96: Same question as mentioned above

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,318 @@
...
+                    try {
+                        LocalDateTime dateTime = convertToDateTime(date);
+                        Deadline deadline = new Deadline(0, desc, dateTime);
+                        tasks.addTask(deadline);
+                    } catch (DateTimeParseException e) {
+                        ui.showInvalidDateFormat();
+                    }
+                }
+            }
+        }
```
antonTan96: This part is quite heavily nested. Could you try to refactor the code to make it more flatter?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,318 @@
...
+                                tasks.addTask(event);
+                            } catch (DateTimeParseException e) {
+                                ui.showInvalidDateFormat();
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
antonTan96: Same question with this method.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public void updateFile(ArrayList<Task> list) {
+        try {
+            //check if file exists, else create
+            File dataFile = new File("./data/duke.txt");
+            if (!dataFile.exists()) {
+                dataFile = createDataFile();
+            }
+
+            //create a FileWriter object to write to file. Note that this overwrites the existing data!
```
antonTan96: I like this warning for i have fallen for this so many times D:
I still don't know how FileWriters work

antonTan96 (commented):
LGTM! There could be some parts that can be optimized but it's a great effort overall

PR #527, opened 2023-09-03
[lordidiot] iP

lordidiot:
# A heading

* A
* bullet
* list

1. A
2. numbered
3. list

```python
def a_fenced_code_block():
    with syntax_highlighting() as x:
        pass
```

- [x] add a task list
- [ ] watch lecture

🤓 

> a blockquote

[a hyper link](https://github.com/nus-cs2103-AY2324S1/ip/pull/527)

`inline code?`

some text formatting: **bold**, *italic*, ~~strikethrough~~ etc.


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+
+        ui.exit();
+    }
+
+    private static boolean handleInput(Parser parser, TaskList taskList, Ui ui) throws DukeException {
+        Task task;
+        switch (parser.getCommand()) {
+        case "bye":
```
Respirayson, line 43: perhaps it would be better to consider the case that the user might input the correct word but a different capitalisation? For eg. instead of bye, the user inputs Bye or BYE.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,116 @@
...
+                isRunning = Duke.handleInput(parser, taskList, ui);
+            } catch (DukeException e) {
+                ui.printException(e);
+            }
+        }
+
+        ui.exit();
+    }
+
+    private static boolean handleInput(Parser parser, TaskList taskList, Ui ui) throws DukeException {
```
Respirayson: I like the function in that it is succinct and it is very readable. However, I think you might want to consider refactoring the code and extracting this function into the Parser class as that class deals with making sense of the user commands.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,114 @@
...
+    public Ui(String name) {
+        this.name = name;
+        this.scanner = new Scanner(System.in);
+    }
+
+    /**
+     * Initialise UI. Prints banner.
+     */
+    public void init() {
+        System.out.println("____________________________________________________________");
```
Respirayson: Small thing but perhaps it would be better to turn the underline into a constant so it can be easily reused elsewhere

src/main/java/duke/TaskStorage.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public ArrayList<Task> loadExistingTasks() throws DukeException {
+        try {
+            if (!this.file.exists()) {
+                this.file.getParentFile().mkdirs();
+                this.file.createNewFile();
+                return new ArrayList<>();
+            }
+
+            ArrayList<Task> tasks = new ArrayList<>();
```
Respirayson: I like how you used tasks as the variable name for the Java Collections Object

src/test/java/duke/mocks/TaskMock.java
```diff
@@ -0,0 +1,21 @@
+package duke.mocks;
+
+import duke.Task;
+public class TaskMock extends Task {
```
Respirayson, line 8: I really like the usage of stubs when performing your unit testing. Really shows you went the extra mile!

src/main/java/duke/TaskStorage.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+
+    /**
+     * Stores serialized tasks in text file.
+     * @param tasks Tasks to store
+     * @throws DukeException Any errors encountered while storing
+     */
+    public void storeTasks(List<Task> tasks) throws DukeException {
+        String serialized = tasks.stream()
```
Respirayson: Very cool usage of streams 👍

Respirayson (commented):
Looks good 👍
Just a few nitpicks here and there

src/main/java/duke/Ui.java
lordidiot: Thank you! That makes sense, will do so.

src/main/java/duke/Duke.java
lordidiot: Definitely, will work on that.

PR #532, opened 2023-09-03
[Cloud7050] iP

Cloud7050:
### Contrived Description

This PR is for the purpose of **receiving peer reviews** and demonstrating knowledge of [**GFMD**](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax). The latter has numerous requirements imposed, therefore, you may find this description to be:
- Highly verbose
- Oddly formatted
- Excessively long-winded

It's a bit like those stereotypical Bollywood films where the editor asks the director how many free transitions they want the movie to have, and the director replies `yes`.

1. 
1. I can't find the main one I wanted to embed, but iirc it has people staring at each other out their car windows with a ton of cuts/transitions/effects applied back to back.

So... we use every syntax in the book. Kinda almost.

Now. Let's change topics so fast that you get literary whiplash. Here is how to ping your router on a loop when it starts acting up:

```bat
ping 192.168.0.1 -t -l 65500
```

And finally, allow me to end with a totally unrelated quote from 📖 "The Art of Computer Game Design" by Chris Crawford:

> Actions that are simple and obvious with other technologies become arcane with the computer. If I give you a bat and tell you that your goal in baseball is to hit the ball, you will have few problems deciding that you should swing the bat at the ball. A computer baseball game is not so easy to figure out. Do you press H for "hit" or S for "swing" or B for "bat"? Do you press the START key or press the joystick trigger? _Perhaps you should swing the joystick by its cable at the ball displayed on the television screen._

---

- [x] a heading
- [x] a bullet list
- [x] a numbered list
- [x] a task list
- [x] an emoji
- [x] a blockquote
- [x] a hyperlink
- [x] inline code
- [x] some text formatting: bold, italic, strikethrough etc.

--- 

I could've just copy-pasted the template, I don't know why I spent so long making this. On the off chance anyone happens by this unremarkable PR... Hope you had a chuckle.


src/com/cloud/chatbot/token/CommandManager.java
```diff
@@ -0,0 +1,90 @@
...
+ * The command is the head, while the details are the tail.
+ *
+ * A flag set is a flag and its sub input that have been processed. Eg "/by
+ * Tomorrow night" split into tokens is a flag set, where the flag is "/by"
+ * whose flag text is "by", and the sub input is "Tomorrow night".
+ */
+public class CommandManager extends TokenManager {
+    // Index by flag text instead of Token for faster retrieval
+    private HashMap<String, FlagManager> flagSets = new HashMap<>();
+
```
AndrewJanong: Shouldn't there be a comment header here before the constructor?

src/com/cloud/chatbot/token/TokenManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * A general parent Token manager providing children useful shared methods.
+ */
+public abstract class TokenManager {
+    protected List<Token> tokens = new ArrayList<>();
+
+    public TokenManager() {}
+
+    protected static List<Token> stringToTokens(String s) {
```
AndrewJanong: Perhaps could use a verb for the function name? Something like `getTokensFromString`

src/com/cloud/chatbot/Cloud.java
```diff
@@ -0,0 +1,196 @@
...
+        case UNKNOWN:
+        default: {
+            Cloud.say(
+                String.format(
+                    "\"%s\" is not a valid command.",
+                    command
+                )
+            );
+            break;
+        }
```
AndrewJanong: Good job on following the coding standard for switch statements 👍

src/com/cloud/chatbot/Cloud.java
```diff
@@ -0,0 +1,196 @@
...
+                    e.getFlagText()
+                )
+            );
+            return null;
+        }
+
+        return new Todo(description);
+    }
+
+    private static @Nullable Integer verifyNumber(CommandManager manager) {
```
AndrewJanong: This function name does not suit with the return type. `verifyNumber` sounds like it would return a `boolean`.

AndrewJanong (commented):
## Overall Review
Overall, good job on following all code quality guidelines for the most part. I like how you keep things simple but still clear and readable. Do take a look at some of the things I pointed out. Keep it up 👍

src/com/cloud/chatbot/Cloud.java
```diff
@@ -0,0 +1,196 @@
...
+import com.cloud.chatbot.token.FlagManager;
+import com.cloud.chatbot.token.Token;
+
+
+
+/**
+ * The chatbot's main class.
+ */
+public final class Cloud {
+    private static final Scanner SCANNER = new Scanner(System.in);
```
siqirua: I like how you capitalize the static variable names

src/com/cloud/chatbot/Cloud.java
```diff
@@ -0,0 +1,196 @@
...
+
+        if (!numberToken.isValidNumber(Cloud.TODOS)) {
+            Cloud.say(
+                String.format(
+                    "TODO #%d does not exist.",
+                    numberToken.toInt()
+                )
+            );
+            return null;
+        }
```
siqirua: I like how you break the line properly to improve readability!

src/com/cloud/chatbot/token/CommandManager.java
```diff
@@ -0,0 +1,90 @@
...
+        this.tokens = TokenManager.stringToTokens(input);
+
+        // Go through tokens to extract flag sets
+        int i = 0;
+        while (i < this.tokens.size()) {
+            Token token = this.tokens.get(i);
+
+            if (token.isFlag()) {
+                int endIndex = this.findFlagSetEnd(i);
+                List<Token> flagSet = this.removeTokens(i, endIndex);
```
siqirua:
```suggestion
                List<Token> flags = this.removeTokens(i, endIndex);
```
Is it better to name this variable as flags or flagList instead of flagSet since one might be confusing about the type of this variable (Set vs List)?

src/com/cloud/chatbot/token/TokenManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+        if (word.length() > 0) {
+            tokens.add(new Token(word));
+        }
+
+        return tokens;
+    }
+
+    protected static String tokensToString(List<Token> tokens) {
+        String joined = "";
```
siqirua: Maybe can use a more meaningful name? It is a bit hard to infer what it is about simply by the variable name.

src/com/cloud/chatbot/todo/Event.java
```diff
@@ -0,0 +1,35 @@
+package com.cloud.chatbot.todo;
+
+
+
+/**
+ * Represents an event TODO, which has both a starting and ending timestamp.
+ */
+public class Event extends Deadline {
+    private String timestampStart;
```
siqirua:
```suggestion
    private String startingTimestamp;
```
Would it make more sense to name the variable as startingTimestamp so it sounds more like a noun?

siqirua (commented):
Overall, the code is easy to read and understand and I like how you follow the coding standards, especially the clear line break and method name. However, though the review should be focusing on the coding standard (so you may ignore this part if other aspects of the code is subject to change), maybe you can consider make a general Task interface and let Todo, Deadline and Event to be three parallel child class? For me, it does not really make sense that a event is also a deadline.

Overall good job! Keep it up!

src/com/cloud/chatbot/token/CommandManager.java
Cloud7050: Ah right, I'd forgotten that after the comment above the class itself. The constructor takes in a param which should be documented. I'll add that in subsequent commits.

src/com/cloud/chatbot/token/TokenManager.java
Cloud7050: I get where you're coming from, though this is meant to be more of a utility function, thus it being a static method. I think names of the form "x to y" also see use elsewhere.

src/com/cloud/chatbot/token/CommandManager.java
Cloud7050: I see what you mean. I hadn't considered the potential mixup with actual `Set`s. It's actually more so referring to this definition:

https://github.com/nus-cs2103-AY2324S1/ip/blob/abb83ad37dc5a8c4d916f4b414881e9e3ed8b10d/src/com/cloud/chatbot/token/CommandManager.java#L16-L18

Since I differentiate between flags and flag sets, the `List<Token>` in question is not really a flag or flags. Instead I prefer to call it a flag set (which includes 1 flag), and is related to the method call above of `.findFlagSetEnd()`.

src/com/cloud/chatbot/token/TokenManager.java
Cloud7050: The string starts empty but gets concatenated onto in subsequent lines, eventually being returned when done. In essence, it is a changing or accumulating temporary variable. If the logic gets more complex, I will consider changing the variable name or adding comments, maybe something like `joinedOutput` or documenting the protected method.

Cloud7050:
> maybe you can consider make a general Task interface and let Todo, Deadline and Event to be three parallel child class? For me, it does not really make sense that a event is also a deadline.

I suppose there are two ways to look at the inheritance.

There's the purely functional view, which I noticed and jumped on before actually scrolling down and seeing the `A-Inheritance` collapse further below the instructions (that talked about `Task` and inheritance). So by the time I read that part, I'd already implemented it "my way" and figured it was ok to stick with. From a properties point of view, `Todo` has no timestamps, `Deadline` has an end, and `Event` has both a start and an end. Due to the properties being a subset of one another, I went for not duplicating them.

Then there's the more real world interpretation, which I totally get. Yes, it does not really make sense for an event to strictly *be* a deadline. Though the way I ended up taking it was that they both have an ending timestamp.

To be honest, I'm not sure which is better. I do like how `Task` is an interface. Maybe it would even fit to have another subinterface like `EndableTask`. So as the project requirements change, I will keep this in mind and take it into account when touching these parts of the code down the line.

@siqirua

src/com/cloud/chatbot/todo/Event.java
Cloud7050: Although I note that those aren't constants, I was actually going for this rule:

image

Ie `timestampStart` matching `timestampEnd`. I've experimented with similar naming schemes elsewhere. I guess I still haven't fully decided on which I prefer more.

I will keep this in mind, though I think once I update the type away from `String`, the new type will make it clear by its nature that those variables are meant to represent a timestamp, and I can just name the variables something like `start` and `end`.

src/com/cloud/chatbot/Cloud.java
Cloud7050: I see what you mean. This one is also more of a utility function. I'm a bit apprehensive about starting static methods with `get` since it sounds to me like an instance getter.

I named it so because part of its logic involves giving error feedback to the user, but at the end of the call, if the number is valid, then the callee needs to actually use said number (while also knowing whether or not it was valid), thus `@Nullable Integer` instead of just `int`.

I am open to better name suggestions.

Cloud7050: Come to think of it, if the method were meant to return a boolean, I think it would be called something like `isNumberValid()`. That thought occurred to me when reading this part of the code quality guide:

image

PR #555, opened 2023-09-04
[H1410101] iP

H1410101:
_For class reasons; ignore._

# CatBot

You might be wondering, why would you use CatBot?

- it has cats

1. There is no better reason, and there needs no other reason.

It features peak programmer syntax, like [this amazing, lovable ~~monstrosity~~ little guy](https://github.com/H1410101/ip/blob/master/src/main/java/catbot/CatBot.java). 

Here's a snippet for your convenience:
```java
BiConsumer<String, BiFunction<
                NamedParameterMap, BiConsumer<ErrorIndicatorIo.InvalidArgumentState, NamedParameterMap>,
                Optional<Task>>>
                createTaskIfValidElseWarn = (args, bifunction) -> slashPattern.ifParsableElseDefault(args,
                        namedParameterMap -> bifunction.apply(
                                namedParameterMap,
                                io::indicateArgumentInvalid
                        ).ifPresent(task -> {
                            taskList.addTask(task);
                            io.displayTaskAdded(taskList);
                        })
                );

        commands.addCommand("todo",
                        args -> createTaskIfValidElseWarn.accept(args, Task.Todo::createIfValidElse)
                )
                .addCommand("event",
                        args -> createTaskIfValidElseWarn.accept(args, Task.Event::createIfValidElse)
                )
                .addCommand("deadline",
                        args -> createTaskIfValidElseWarn.accept(args, Task.Deadline::createIfValidElse)
            );
```

**That's right!** Your :eyes: do not deceive you. That is a 
- [x] `BiConsumer` that takes in a 
- [x] `BiFunction`, which takes in a 
- [x] `BiConsumer` as one of its arguments
- [x] AND returns an `Optional`! 

I bet you've never seen that many angle brackets in a row since CS2030S.

> Programs must be written for machines to execute, and only incidentally for people to partially comprehend in flashes of brilliance after large swathes of agony. 
~ Me, 2023, while descending into the infinite abyss, 2 weeks behind schedule


src/main/java/catbot/CatBot.java
```diff
@@ -0,0 +1,102 @@
...
+                )
+                .addCommand("event",
+                        args -> createTaskIfValidElseWarn.accept(args, Task.Event::createIfValidElse)
+                )
+                .addCommand("deadline",
+                        args -> createTaskIfValidElseWarn.accept(args, Task.Deadline::createIfValidElse)
+                )
+        ;
+
+    }
```
rayyan35p: Perhaps you could have added this to another class like CommandMap?

src/main/java/catbot/CatBotCommandPatterns.java
```diff
@@ -0,0 +1,63 @@
+package catbot;
+
+import catbot.internal.CommandPattern;
+import catbot.internal.CommandPatternGenerator;
+import catbot.internal.NamedParameterMap;
+
+import java.util.function.Consumer;
+
+public class CatBotCommandPatterns {
```
rayyan35p: Maybe you can consider adding header comments/javadocs? For all your other classes as well

src/main/java/catbot/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+
+    //region Internal Helpers
+
+    /**
+     * Split string into one pair of command + argument based on the first whitespace.
+     * @param s string containing both command and argument, in that order
+     * @param map to store the mapping between command and argument
+     */
+    private static void parseCmdArg(String s, NamedParameterMap map) {
+        String[] cmdArg = s.split("\\s", 2);
```
rayyan35p: Perhaps you can write out the variable name of `cmdArg` to be in full so that it is clearer?

src/main/java/catbot/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    private static void parseCmdArg(String s, NamedParameterMap map, boolean emptyArgument) {
+        if (emptyArgument) {
+            map.addNamedParameter("", s.trim());
+        } else {
+            parseCmdArg(s, map);
+        }
+    }
```
rayyan35p: I like that you still wrapped in curly brackets even though they are single statement!

rayyan35p (commented):
Good code quality mostly, just need to add header comments and rename variables to make it clear. 💯

src/main/java/catbot/internal/NamedParameterMap.java
```diff
@@ -1,11 +1,37 @@
...
-public class NamedParameterMap extends HashMap<String, String> {
+public class NamedParameterMap {
+
+    private HashMap<String, String> parameters;
+
+    public NamedParameterMap() {
+        parameters = new HashMap<>();
+    }
+
+    public boolean containsKey(String key) {
```
remuslum, line 14: can consider renaming method to avoid confusion

remuslum (commented):
Good job 💯

src/main/java/catbot/CatBotCommandPatterns.java
H1410101: Yup! I'll get around to it. _Eventually_. :')

src/main/java/catbot/CatBot.java
H1410101: I actually spent quite a bit deliberating this, but I ended up not doing it (at least for now).

## The part where I defend myself
My thought process was that I wanted to add bot-specific things in bot-specific classes; to change what commands are supported and what behaviours result from commands, my current iteration only requires that I change or even replace `CatBot` (and possibly `CatBotConsoleIO`) with minimal rewrite.
While there might be some merit in a `CatBotCommands` class that contains this method, that would leave CatBot as a small, anaemic class (I am leaning towards this being a bad thing, but I'm not too sure of myself here).
Ultimately, I had decided that it made more sense to me to have the commands be clear from the bot description, instead of its own separate class, especially since the class would otherwise be so small as to be trivial.

## The part where I stop defending myself
I definitely see an alternative that takes your considerations into account. I could rename the current file to `CatBotMain` or `CatBotEntrypoint` to reflect its explicit purpose as a small and trivial class, put this initialization as a protected static method of `CatBotCommandLoader`, or have it be the default initialization of `CatBotCommandMap`. 
I'd love to hear your thoughts on why this design might be preferable, though; or if you had a different design in mind, I'd like to chew on that as well.

Regardless, thanks for the feedback!

src/main/java/catbot/Parser.java
H1410101: Good catch, will do. Thanks!

src/main/java/catbot/CatBot.java
```diff
@@ -0,0 +1,102 @@
...
+                )
+                .addCommand("delete",
+                        string -> runIfValidIndexElseIndicateError.accept(string,
+                                validIndex -> io.printTaskDeleted(taskList.removeTask(validIndex-1))
+                        )
+                );
+
+        // User creating new tasks (with SlashPattern)
+
+        BiConsumer<String, BiFunction<NamedParameterMap, BiConsumer<ErrorIndicatorIo.InvalidState, NamedParameterMap>, Optional<Task>>>
```
nubnubyas: Perhaps u can split this line into 2? I think this line has 135 characters which exceeds the 120 characters limit stated in the coding standard.

src/main/java/catbot/io/CatbotConsoleIO.java
```diff
@@ -0,0 +1,229 @@
...
+    public String getCommand() {
+        if (this.endOfCommand == null) findCommand();
+        if (this.endOfCommand == -1) return retrieve();
+        else return retrieve().substring(0, this.endOfCommand);
+    }
+
+    public String getArgs() {
+        if (this.endOfCommand == null) findCommand();
+        if (this.endOfCommand == -1) return "";
+        else return retrieve().substring(this.endOfCommand+1);
```
nubnubyas: Perhaps it would be better to have the conditionals on a separate line. E.g
if (this.endOfCommand == null) {
    findCommand();
}

src/main/java/catbot/io/CatbotConsoleIO.java
```diff
@@ -0,0 +1,229 @@
...
+        return lastUserInput;
+    }
+
+    public String retrieve() {
+        return lastUserInput;
+    }
+
+    private void findCommand() {
+        if (retrieve() != null)
+            this.endOfCommand = retrieve().indexOf(" ");
```
nubnubyas: It would be better to have curly brackets here based on the java coding standard.

nubnubyas (commented):
Good job for the chatbot! I really like your usage of java.util.function in your code. Just some small coding standard problems here and there. Overall, I really like your IP.

src/main/java/catbot/CatBot.java
```diff
@@ -0,0 +1,102 @@
...
+    static {
+        addSupportedCommandsToCommandMap();
+    }
+
+    private static void addSupportedCommandsToCommandMap() {
+        CommandPattern<Integer> integerPattern = CatBotCommandPatterns.getIntegerPatternGenerator()
+                .generate(io::indicateInvalidInteger);
+        CommandPattern<NamedParameterMap> slashPattern = CatBotCommandPatterns.getSlashPatternGenerator()
+                .generate(CatBotCommandPatterns.NO_DEFAULT);
+
```
btaanish: I like how you name the variables in a greater explanatory sense

src/main/java/catbot/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+    private Parser(String delimiter, boolean emptyArgument) {
+        this.delimiter = delimiter;
+        this.emptyArgument = emptyArgument;
+    }
+
+    public static Parser by(String delimiter) {
+        return by(delimiter, false);
+    }
+
+    public static Parser by(String delimiter, boolean emptyArgument) {
```
btaanish: Perhaps adding javadocs for public functions might help explain their functioning in greater detail?

src/main/java/catbot/CatBotCommandPatterns.java
```diff
@@ -0,0 +1,63 @@
...
+                        invalidInput.accept(args);
+                    }
+                }
+            };
+        }
+    }
+    //endregion
+
+    //region Slash Arguments Pattern
+    private static final SlashArgumentPatternGenerator slashPatternGenerator = new SlashArgumentPatternGenerator();
```
btaanish: Perhaps having shorter variable names may increase the readability of the code?

src/main/java/catbot/io/CatbotConsoleIO.java
```diff
@@ -0,0 +1,229 @@
...
+    //endregion
+
+    //region Internal Helper
+
+    private String format(String s, String prefix) {
+        if (s == null) return "";
+        return s.replaceAll("(^|\n)", "$1"+prefix);
+    }
+
+    private void line() {
```
btaanish: Perhaps adding a verb in the function name might be more intuitive?

src/main/java/catbot/task/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+        update();
+    }
+
+    public Task removeTask(int index) {
+        Task removed = tasks.remove(index);
+        update();
+        return removed;
+    }
+
+    public void ifValidIndexElse(int index, Consumer<Integer> ifValid, Consumer<Integer> otherwise) {
```
btaanish: Perhaps adding context to the variable names might communicate their purpose better?

btaanish (commented):
Overall a nice attempt with a close adherence to the coding standars. Just a few nits here and there but generally well designed.

src/main/java/catbot/CatBot.java
H1410101: Got it, and thanks for all the nudges for coding style. 
I ended up running it through a stylechecker (finally), but I appreciate the pointers!

src/main/java/catbot/Parser.java
H1410101: Noted! I'll probably get back to it at some point.

src/main/java/catbot/io/CatbotConsoleIO.java
H1410101: Good catch; I'll do that!

src/main/java/catbot/CatBotCommandPatterns.java
H1410101: Perhaps, but these are very infrequently used, so I thought it might have been better to make it very descriptive instead.
I'll think about it, but as of right now I'm still leaning towards my current convention, even if it results in excessively long variable names.

src/main/java/catbot/task/TaskList.java
H1410101: This is heavily inspired by `Optional`'s [`ifPresentOrElse`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#ifPresentOrElse(java.util.function.Consumer,java.lang.Runnable)) (and other similar methods), and I'm trying to take it to its logical extensions; if, even with context, this does not seem intuitive enough, I'm happy to discuss about how to make it better!

PR #557, opened 2023-09-04
[Tyrus Lye]iP

TyrusLye:


src/main/java/Duke.java
```diff
@@ -39,10 +42,17 @@ public static void main(String[] args) {
                     System.out.println("Got it. I've added this task:\n  " + tasks.get(tasks.size() - 1));
                     System.out.println("Now you have " + tasks.size() + " tasks in the list.");
                 } else if (command.startsWith("deadline")) {
-                    String description = command.substring(9, command.indexOf("/by")).trim();
-                    String by = command.substring(command.indexOf("/by") + 4).trim();
-                    tasks.add(new Deadline(description, by));
+                    // Parse the date and time in the format d/M/yyyy HHmm
+                    String[] parts = command.split(" /by ");
+                    if (parts.length < 2) {
```
remuslum, line 47: Suggestion: Provide an example of a suitable date format to enter e.g. 17/4/2023 2PM

src/main/java/Duke.java
```diff
@@ -3,22 +3,21 @@
...
-        ArrayList<Task> tasks = loadTasksFromFile();
+        ArrayList<Task> tasks = new ArrayList<>(100);
+        loadTasksFromFile(tasks);
         String logo = "██╗░░░██╗██████╗░██████╗░░█████╗░██╗\n"
-                    + "██║░░░██║██╔══██╗██╔══██╗██╔══██╗██║\n"
-                    + "██║░░░██║██████╔╝██████╦╝██║░░██║██║\n"
-                    + "██║░░░██║██╔══██╗██╔══██╗██║░░██║██║\n"
-                    + "╚██████╔╝██║░░██║██████╦╝╚█████╔╝██║\n"
-                    + "░╚═════╝░╚═╝░░╚═╝╚═════╝░░╚════╝░╚═╝\n";
+                + "██║░░░██║██╔══██╗██╔══██╗██╔══██╗██║\n"
```
remuslum, line 15: Nice logo 👍🏻

remuslum (commented):
Good job, keep it up 💯

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+                + "░╚═════╝░╚═╝░░╚═╝╚═════╝░░╚════╝░╚═╝\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
+            String command = scanner.nextLine();
+            System.out.println("____________________________________________________________");
+            try{
+                if (command.equalsIgnoreCase("bye")) {
```
aarontxz: would it be better to use enumerates for the commands?

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+                    Task task = createTaskFromLine(line);
+                    if (task != null) {
+                        tasks.add(task);
+                    }
+                }
+                fileScanner.close();
+            }
+        } catch (FileNotFoundException e) {
+            // Handle file not found exception
+            System.out.println("File not found: " + FILE_PATH);
```
aarontxz: I like that the handling of this exception not only tells that user that the file cannot be found but also returns the filepath

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+            try{
+                if (command.equalsIgnoreCase("bye")) {
+                    System.out.println("Bye. Hope to see you again soon!");
+                    break;
+                } else if (command.equalsIgnoreCase("list")) {
+                    System.out.println("Here are the tasks in your list:");
+                    for (int i = 0; i < tasks.size(); i++) {
+                        System.out.println((i + 1) + ". " + tasks.get(i));
+                    }
+                } else if (command.startsWith("todo")) {
```
aarontxz: would it be better to package some of these if, else if else steps into a function that handles everything within the outer most if case? for eg. a function to handle creating todo here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+                    tasks.add(new Event(description, from, to));
+                    System.out.println("Got it. I've added this task:\n  " + tasks.get(tasks.size() - 1));
+                    System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                } else if (command.startsWith("mark")) {
+                    int index = Integer.parseInt(command.split(" ")[1]) - 1;
+                    if (index >= 0 && index < tasks.size()) {
+                        tasks.get(index).markDone();
+                        System.out.println("Nice! I've marked this task as done:\n  " + tasks.get(index));
+                    } else {
+                        System.out.println("Invalid task index.");
```
aarontxz: would it be useful if instead of printing invalid task index, the size of the tasklist is also returned in this else case to let user know that for example there is only 3 tasks in the tasklist thats why their input 4 is invalid. Something like "Invalid task index, the task list has " + tasklist.size() + "tasks. "

aarontxz (commented):
Code seems logical and can handle tasks given, could package more chunks of codes into functions to make the code neater. Some quality of life changes for the user are commented too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+        System.out.println("____________________________________________________________");
+
+        while (true) {
+            String command = scanner.nextLine();
+            System.out.println("____________________________________________________________");
+            try{
+                if (command.equalsIgnoreCase("bye")) {
+                    System.out.println("Bye. Hope to see you again soon!");
+                    break;
+                } else if (command.equalsIgnoreCase("list")) {
```
nananakx-x: Maybe it will be better if this part is separated into different functions like addToDoTask, etc. to improve readability, what do you think?

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+        if (parts.length < 3) {
+            return null;
+        }
+
+        String type = parts[0];
+        String status = parts[1];
+        String description = parts[2];
+
+        Task task = null;
+        switch (type) {
```
nananakx-x: For switch statements, I think there shouldn't be indentation on the next line. On the mod website:
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
+    public Todo(String description) {
+        super(description,null);
+    }
+    @Override
+    public String toFileString() {
+        return "T | " + (isDone ? "1" : "0") + " | " + description;
+    }
+
+    @Override
+    public String toString() {
```
nananakx-x: Clean and concise, good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,272 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>(100);
+        loadTasksFromFile(tasks);
+        String logo = "██╗░░░██╗██████╗░██████╗░░█████╗░██╗\n"
```
nananakx-x: This is cool!!

nananakx-x (commented):
LGTM overall!! The naming of the methods are all of the course standard, just that it might be better for you to start adding javadoc to your public methods and also take note of the indentation of switch statements. Readability can also be improved if commands are split into different methods to handle each case (instead of putting all in one method). Good job so far! :D

PR #568, opened 2023-09-07
Branch level 10

lululwtv:


src/main/java/OUTPUT.txt
```diff
@@ -0,0 +1,5 @@
+[T] [X] cock
```
LinWanLeii, line 1: Really love the test cases mentioned here, just personal preference, if the task type and Marked is placed side by side it would be nicer

src/main/resources/images/idol.png
LinWanLeii, line 1: Awesome use of photo i love it.

src/main/resources/images/chad.png
LinWanLeii, line 1: looks like me in real life so very nice.

LinWanLeii (commented):
Honestly very nicely done, i got nothing much to review on good job.

PR #571, opened 2023-09-07
[Luah Jun Yang} iP

LuahJunYang:
Frenchie keeps track of all tasks in your life in a fuss-free CLI. It is:
- Simple to learn
- Text-commands based
- **SUPER FAST**

All you need to start using Frenchie is:

1. Download Frenchie from [here](https://github.com/LuahJunYang/ip)
2. Double click the jar file
3. Enter your tasks in the right format
4. Let it track all of your tasks for you 👍🏻 

Features:

- [x] Task Tracking
- [ ] _Reminders for Tasks with deadlines (coming soon!)_

For those of you who code, Frenchie is written in java and you can experiment with it to add your own features!

Here is the main() method of Frenchie:
```java
        new Frenchie("frenchie.txt").run();
    }
```


src/main/java/Frenchie.java
```diff
@@ -79,6 +81,16 @@ public static void main(String[] args) {
...
+                } else if (input.contains("delete")){
+                    String[] parts = input.split(" ");
+                    int index = Integer.parseInt(parts[1]) - 1;
+                    Task target_task = frenchie.tasks.get(index);
+                    frenchie.deleteTask(index);
+                    System.out.println("____________________________________________________________\n" +
+                            "Noted. I've removed this task: \n" +
+                            target_task.toString()   + "\n" +
+                            "Now you have " + frenchie.getNumOfTasks() + " tasks in the list.\n" +
+                            "____________________________________________________________");
```
sushiyade: As the number of keywords increase, perhaps you might want to start looking at enums to have a collection of your keywords!

src/main/java/Frenchie.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class Frenchie {
+    public ArrayList<Task> tasks;
+
+    //constructor
+    public Frenchie() {
+        this.tasks = new ArrayList<>();
+    }
+
+    public void addTask(Task NEW_TASK) {
```
yytan25: The Task parameter is a variable, not a constant. So it should be in camelCase instead, i.e. `newTask`.

src/main/java/Frenchie.java
```diff
@@ -0,0 +1,144 @@
...
+    public void listTasks() {
+        int counter = 1;
+        for (Task task : tasks) {
+            System.out.println(counter + ". " + task.toString());
+            counter += 1;
+        }
+    }
+
+    public void completeTask(int index) {
+        tasks.get(index).mark_as_completed();
```
yytan25: Please follow the camelCase convention for methods, i.e. `markAsCompleted`.

src/main/java/Frenchie.java
yytan25: Seconded. In addition, perhaps you may wish to divide all the logic in your `main` method into several other methods, as the complexity of your program increases.

src/main/java/Frenchie.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+    public static void main(String[] args) {
+        /*String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo); */
+        Frenchie frenchie = new Frenchie();
+        String skeleton = "____________________________________________________________\n" +
```
yytan25: When breaking up a statement into multiple lines, the `+` operator should be on a new line.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,11 @@
+class ToDo extends Task {
```
yytan25: Perhaps you should add the `public` access modifier to your class definition?

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task{
```
yytan25: Should leave a whitespace between your definition and the curly braces, as in
`public class Event extends Task {`

src/main/java/Frenchie.java
```diff
@@ -0,0 +1,144 @@
...
+            String input = scanner.nextLine();
+            try {
+                if (input.equals("bye")) {
+                    System.out.println("____________________________________________________________\n" +
+                            " Bye. Hope to see you again soon!\n" +
+                            "____________________________________________________________");
+                    break;
+                } else if (input.equals("list")) { //Checking if user is looking to list all tasks
+                    frenchie.listTasks();
+                } else if (input.contains("mark")) { //Checking if user input is to mark/unmark tasks
```
yytan25: This doesn't check if the input is either `mark` or `unmark`. If a user inputs `markzzzzzz`, this will still be treated as `unmark` even though it is an invalid command, because the input still contains the substring `mark`. To avoid similar problems like this, perhaps you should use Java Enums to ensure input validity.

yytan25 (commented):
Good use of OOP and inheritance for the subclasses of `Task`. You can still improve on your coding style and your use of SLAP.

src/main/java/Frenchie/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+                } else if (taskType.equals("E")) {
+                    String taskName = taskDetails.split("\\(")[0].trim();
+                    String startTime = taskDetails.split("\\(")[1].split("from: ")[1].split(" to")[0];
+                    String endTime = taskDetails.split("\\(")[1].split("to: ")[1].split("\\)")[0];
+                    DateTimeFormatter inputFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
+                    LocalDateTime storedStartTime = LocalDateTime.parse(startTime, inputFormatter);
+                    LocalDateTime storedEndTime = LocalDateTime.parse(endTime, inputFormatter);
+                    DateTimeFormatter desiredFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm");
+                    String constructorStartTime = storedStartTime.format(desiredFormatter);
+                    String constructorEndTime = storedEndTime.format(desiredFormatter);
```
Kevin-Liusx, line 53: I think this part of your code and be improved? Maybe try to use some other methods?😊

